PR #27, opened 2015-07-15
Get geojs from github

mgrauer:
@aashish24 This is ready for review

I've set the geojs path in package.json to pull from github, which will occur when `npm install` is run at the top level of minerva.

The problem is that this doesn't do a full clone of geojs, it doesn't include the `.git` directory, which is why I was getting silent failures on the `git submodule init` and `git submodule update` calls that I mentioned before.

So I can either 

1) Accept the `geo.min.js` that is built into the project and downloaded when I perform `npm install`.  The problem here is that the `geo.min.js` is fairly out of date as it hasn't been updated for a while.

or

2) Delete the `vgl` subdir, clone `vgl`, update `vgl` to the correct commit sha, then `npm install` and `grunt` within the geojs dir to build a current version of geojs

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,3 +51,20 @@ Post-Minerva:
 
 http://localhost:8080           => serves Minerva
 http://localhost:8080/girder    => serves Girder
+
+
+#### Installing a specific version of GeoJs for development
+
+This is to get around the fact that npm is for installing packages, not managing source repositories.  So when npm installs geojs, it doesn't install it as a git repo with the .git dir.
+
+  1. change the reference in the `geojs` line of package.json to point to the correct branch or sha reference
+  2. run `npm install` from the top level `minerva` directory
+  3. `cd node_modules/geojs`
+  4. `rm -rf vgl`
+  5. `git clone https://github.com/OpenGeoscience/vgl.git`
+  6. find the hash of vgl from the geojs branch
+  7. `cd vgl; git checkout $HASH_FROM_GEOJS_BRANCH; cd ..`
+  8. `npm install` (this time inside geojs)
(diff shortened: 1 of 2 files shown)


aashish24:
Why don't we update the geo.min.js? Unless it is pointing to a release version.

jbeezley:
Hmmm... Another problem with using git submodules.  The standard way of dealing with this would be to publish vgl on npm rather than include it as a submodule.

`geo.min.js` is the most recent release.  I don't want to update that file with every PR because it is a single line.  It would add a 300K blob to the diff every time.  Unfortunately, we can't add a submodule update to the `npm install` step (an issue to do so was [rejected](https://github.com/npm/npm/issues/6700) for good reason), so there will always need to be some custom logic in minerva if we want to install from a git head.  The only other option is to cut a release of geojs every time we need to update and use the npm version.

mgrauer:
The point of this was to allow development of Minerva using specific branches of GeoJs.  I don't think we want to put much time into getting an automated solution.  Here's what I propose:

 * leave the package.json pointing to github/opengeoscience/geojs#master as it is in this commit
 * add some instructions for developers of minerva to manually build vgl/geojs when they want anything other than the current released version of geojs
 * at some future point further into minerva's maturity, once we have actual releases of minerva, tie them to specific released versions of geojs, meaning we repoint package.json to an npm version

aashish24:
@mgrauer +1 for your proposal. @jbeezley I agree and thats what I referred to in my comment.

mgrauer:
Merging, as this satisfies the proposal and doesn't introduce any new code.

PR #28, opened 2015-07-16
Add support for importing an S3 bucket as a dataset

cjh1:


cjh1:
@mgrauer This is ready for review

mgrauer:
can you add the jobs plugin as a dependency in plugin.json?

mgrauer:
I've merged master and pushed that commit.

Is it possible to add server side tests here?  If not, at least the current tests should pass.

cjh1:
@mgrauer Added a testcase ( and fix a bug it revealed! )

mgrauer:
Thanks @cjh1 .  Hooray for testing!

PR #29, opened 2015-07-16
Analysis collection folder

mgrauer:
ready for review

### DIFF
--- a/server/rest/analysis.py
+++ b/server/rest/analysis.py
@@ -1,0 +1,45 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+
+###############################################################################
+#  Copyright Kitware Inc.
+#
+#  Licensed under the Apache License, Version 2.0 ( the "License" );
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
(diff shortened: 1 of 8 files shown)


server/rest/analysis.py
```diff
@@ -0,0 +1,45 @@
...
+        return {'folder': folder}
+    getAnalysisFolder.description = (
+        Description('Get the minerva analysis folder.'))
+
+    @access.user
+    def createAnalysisFolder(self, params):
+        folder = findAnalysisFolder(self.getCurrentUser(), create=True)
+        return {'folder': folder}
+    createAnalysisFolder.description = (
+        Description('Create the minerva analysis folder, a global resource.'))
```
cjh1: I guess we need to consider who will create the analysis folder and probably have some sort of group owner ship so everyone can read the items in it? This is related to how the analyses are imported as well.

cjh1:
LGTM

aashish24:
@mgrauer I will run the test if that's okay before I can approve it..

server/rest/analysis.py
cjh1: I am actually thinking that maybe the collection and folder should be created at deployment time ( using for example a girder client script ) rather than have specific endpoints for this?

mgrauer:
@aashish24 fire away with tests.  We should probably try to get Travis testing this soon.

mgrauer:
@cjh1 I agree with both of your points:

  1. some kind of group for minerva users
  2. deployment time script to create the collection

The endpoints make it easy for the web client to interact with a minerva instance without worrying about whether the proper folders have already been created, and this also makes it easier I think for developers who are going back and forth between different girder based projects, which is why I wanted the endpoints.

cjh1:
@aashish24 I have run the tests and they pass, can we merge this?

aashish24:
Go ahead. I have some questions but I can ask them in person!

aashish24:
+2 LGTM

PR #31, opened 2015-07-17
Add script to import romanesco analysis

cjh1:
@mgrauer I can rebase this when the other branches are in master ...


cjh1:
@mgrauer This ready for review

.travis.yml
```diff
@@ -1,7 +1,6 @@
 language: python
 python:
     - "2.7"
-    - "3.4"
```
mgrauer: I wonder if it's better to comment this out with a note that romanesco is what is preventing us from using 3.4, or maybe better in the docs?

.travis.yml
```diff
@@ -49,6 +48,12 @@ before_install:
     - npm install -g npm@next
     - npm --version
 
+    # Install R staisfy romanesco
```
mgrauer: satisfy?

import_analyses.py
```diff
@@ -0,0 +1,59 @@
...
+    if not analyses_folder['folder']:
+        analyses_folder = client.post('/minerva_analysis/folder')
+
+    analyses_folder =  analyses_folder['folder']
+
+    for analysis_file in glob.glob(os.path.join(analyses_path, "*.json")):
+        analysis = romanesco.load(analysis_file)
+        analysis_name = analysis['name']
+
+        # See if we already have an analysis with with name
```
mgrauer: with with => with that

mgrauer:
@cjh1 what do you think of:

-move the add.json to somewhere under plugin_tests, like plugin_tests/analyses
-add an option to the import_analysis utility to take a folder path, then in the test when you use import_analysis you can point to plugin_tests/analyses/add.json
-move import_analysis to utility/import_analysis where utility is a sibling of server

?

cjh1:
@mgrauer Good plan, I will push that changes

cjh1:
@mgrauer How does this look now?

mgrauer:
Thanks for the fixes.  A problem I noticed is that tests are failing but Travis is passing--probably something I did wrong when I first set Travis up but didn't notice before.  

I think there are (at least) two issues here:

  * Travis should fail when tests fail
  * Romanesco needs to be git cloned in the travis file, unless it is happening in one of those eval steps

I'd like to see the "Travis should fail" one fixed first, so we can see that Travis (correctly) fails while we have failing tests.  Do you want me to fix this and then hand back to you for Rom install?

cjh1:
@mgrauer Sounds good if you can get Travis to fail I will fix up the tests.

mgrauer:
I'll look into Travis.

It's probably slightly more complicated than just cloning Romanesco--as you have it in the pip requirements.txt, but Travis is giving us

    115: Exception: Plugin directory does not exist: /home/travis/build/girder/plugins/romanesco

mgrauer:
True negative worked, I'll remove this WIP from the branch and force push.

mgrauer:
@cjh1 all yours, Travis appears working now with respect to CTest.

cjh1:
@mgrauer I declare victory!

mgrauer:
Nice @cjh1 , way to get the tests passing as true negatives!  :v:

ping @zachmullen also here, to look at this last commit, in case he has feedback or just wants to be aware of this rollout strategy.  We have Romanesco as an enabled plugin in Girder, and install it through pip, so we needed to add a symlink from the plugins/romanesco dir to the actual pip installed romanesco dir.

PR #36, opened 2015-07-18
Run Minerva tests in Travis

### LINKED ISSUE #34: Turn on travis

mgrauer:
closes #34 .

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,61 @@
+language: python
+python:
+    - "2.7"
+    - "3.4"
+
+sudo: false
+
+compiler:
+    - gcc
+
+addons:
+  apt:
+    packages:
+    - libfreetype6-dev
+    - libpng12-dev
+    - libgdal1-dev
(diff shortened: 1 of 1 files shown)


mgrauer:
Travis works! 

@jbeezley A lot of the time spent in the build is installing Gaia and its dependencies.  

@cjh1 has an active branch where he is bringing in Romanesco.  

Let's all meet up this week and talk about Romanesco/Gaia integration, along with testing and dependency management.

jbeezley:
That will be more an more of an issue as we go on.  There are a few options to handle it:

1. Do what I did for romanesco and serve the dependencies as a tarball at data.kitware.com.
2. Use something like [mason](https://github.com/mapbox/mason) which does essentially the same thing but better.  This will probably require a minor change to push builds to girder rather than s3.
3. Use circleci, which allows you persist directories between builds.  I have used it as part of cdatweb testing to cache a very large docker image, and it works pretty well.

PR #37, opened 2015-07-18
Add dependency on jobs plugin

mgrauer:
@cjh1 We forgot to add this into your S3 branch.  I'll go ahead and merge.


cjh1:
@mgrauer Sorry about that, thanks for adding it

mgrauer:
I missed it too @cjh1 .  :sunglasses:

PR #41, opened 2015-07-20
Only use assets from geojs dist directory if it exists

cjh1:


cjh1:
@kotfic @mgrauer This should fix the build issue we discussed

mgrauer:
LGTM.  Merging.

PR #48, opened 2015-07-24
Add iso-contour analysis

cjh1:
This analysis takes a file id for a NetCDF file and converts the file
into a format that can be rendered as an iso-contour layer in GeoJs.
Currently we only convert a small section of the file, so it can be
rendered on the client side.


cjh1:
@mgrauer This is our first real analysis for NEX, I have it running in romanesco

aashish24:
Yay!!

cjh1:
@mgrauer This is ready for review now

mgrauer:
LGTM

PR #51, opened 2015-07-27
Instantiate analysis widgets by name

cjh1:
This should allow for multiple analysis types without having to
update the AnalysisPanel each time.


mgrauer:
LGTM, nice work :leaves: 

our future selves need to figure out how to pass specific params into widgets.

PR #52, opened 2015-07-27
Datasetinfo s3 hierarchy

kotfic:
Implements a hierarchical widget that allows users to select a file from an S3 dataset.  This information is then persisted back to the server in the dataset's 'minerva' metadata field.

### DIFF
--- a/server/rest/s3_dataset.py
+++ b/server/rest/s3_dataset.py
@@ -108,7 +108,8 @@ class S3Dataset(Resource):
             'bucket': bucket,
             'prefix': prefix,
             'assetstoreId': str(target_assetstore['_id']),
-            'folderId': prefix_folder['_id']
+            'folderId': prefix_folder['_id'],
+            'selectedItems': []
         }
 
         metadata['minerva'] = minerva_metadata
--- a/server/rest/s3_dataset.py
+++ b/server/rest/s3_dataset.py
@@ -128,13 +128,13 @@ class S3Dataset(Resource):
         .param('id', 'The Item ID', paramType='path')
         .errorResponse('ID was invalid.'))
 
-
     @access.public
     @loadmodel(model='item', level=AccessType.WRITE)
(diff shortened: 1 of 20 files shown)


server/rest/s3_dataset.py
```diff
@@ -122,3 +127,37 @@ def createDataset(self, item, params):
...
+        params['folderId'] = ObjectId(params['folderId'])
+
+        # Metadata keys we allow to be updated
+        valid_keys = ['selectedItems']
+
+        try:
+            minerva_metadata = item['meta']['minerva']
+        except KeyError:
+            raise RestException(u"No metadata set, is this really a dataset?")
+
```
cjh1: Why use unicode string here?

server/rest/s3_dataset.py
```diff
@@ -122,3 +127,37 @@ def createDataset(self, item, params):
...
+        # Strictly allow only updates to keys in valid_keys
+        if len(set([i for i in minerva_metadata.items() if i[0] not in valid_keys]) ^
+               set([i for i in params.items() if i[0] not in valid_keys])) > 0:
+            raise RestException("Only %s may be updated!" % valid_keys)
+
```
cjh1: Do you think this could be simplified a little, don't we just want to check that all keys are in valid_keys?

kotfic: Probably could be simplified (or better named?),  the idea was to white list values that can be updated through this method and disallow all other metadata values to changed.

keys like assetstoreId and and original_type shouldn't be allowed to change correct?

kotfic:
Locally both pep8 and jsstyle_minerva are passing - not sure why travis is showing failure?

cjh1:
@kotfic Is this ready to merge?

kotfic:
@cjh1 Should be ready assuming tests pass

cjh1:
LGTM

PR #53, opened 2015-07-28
Bsve analysis creation

mgrauer:


analyses/bsve/search/minerva.json
```diff
@@ -0,0 +1 @@
+{ "minerva": { "analysis_type": "bsve_search" } }
```
cjh1: Rather than having a separate file for the analysis type, why don't we just use the analysis name which is already contained in the analysis JSON?

utility/import_analyses.py
```diff
@@ -1,37 +1,57 @@
...
+    # e.g. analyses/bsve
+
+    for analysis_subfolder in os.listdir(analyses_path):
+        analysis_path = os.path.join(analyses_path, analysis_subfolder)
+
+        # If there is an analysis.json, it is a Romanesco analysis
+        romanesco_analysis = os.path.join(analysis_path, 'analysis.json')
+        romanesco_metadata = {}
+        if os.path.exists(romanesco_analysis):
+            analysis = romanesco.load(romanesco_analysis)
```
mgrauer: here's where we load the analysis into Romanesco

analyses/bsve/search/minerva.json
mgrauer: What I was worried about is creating a Romanesco analysis where there isn't one.  e.g., For the bsve search, which is currently a local job, would it be a problem to call romanesco.load?  I pointed out in another comment where we are calling load.

Even if it isn't a problem to call romanesco.load, it is a bit misleading to call that on an analysis that shouldn't be run under romanesco.

I agree generally and would like to remove the extra minerva.json if possible.

mgrauer:
I'll remove the minerva.json file from romanesco analyses.  If there is a romanesco analysis.json, we'll take the `analysis_type` from snake_casing the last two folders containing the analysis.  

If there isn't a romanesco analysis, I'll leave in a minerva.json file, and add an `analysis_name` there.

At first I had thought to fully remove the minerva.json file from all cases, but this doesn't seem explicit enough in cases where there is no romanesco analysis.

cjh1:
LGTM

cjh1:
@mgrauer If the test pass I think we are good to go

PR #54, opened 2015-07-28
Add contour analysis

cjh1:
This should be merge after "generic-analysis-widget-support"

Still todo:

- [ ] Need to validate input
- [x] Need to ensure that the dataset panel is updated after the analysis is complete.
- [ ] Error handling in the case the analysis fails
- [x] Remove hard coded fileId!
- [ ] The analysis in romanesco should use a separate Girder auth token, currently is uses the web clients token.


kotfic:
@mgrauer  @cjh1  Could you please review 303efd1  this makes non-trival changes to the initialization and rendering order of the entire interface.   Thanks!

mgrauer:
@kotfic I like the direction you are taking in 303efd1, it improves the rendering logic and I confirm that it does dynamically update the list of datasets when I run an analysis job that creates a dataset.  You'll want to remove the commented out code as part of the final cleanup for the PR.

For the current implementation of `renderableP`, this does not allow datasets that include shapefiles that have been converted to geojson to be rendered.  Since it assumes that all json is contour json, it shows some json datasets as being renderable when they are not, i.e. when they would need to have some mapping defined and then a geojson conversion run before they could be rendered.  Perhaps this will be cleared up as part of a later pass that splits out different types of Datasets into model subclasses?

kotfic:
@mgrauer yeah,  ```renderableP``` in addition to being unfortunately named,  is really only intended as a stub until a better typing system can be devised. A refactor that introduces dataset specific models (like S3DataModel but for CSV, Mongo, geojson,  etc)  would be preferable -  for now I can put in some additional logic in ```renderableP``` to detect shapefile conversion  -  is there a test shapefile available somewhere?

kotfic:
@mgrauer @cjh1 @aashish24  Also,  contour rendering is working and once OpenGeoscience/geojs#427 is resolved I think we should have everything needed for an end-to-end demo (not including spark) of selected relatively small files. 

I'd like to merge this branch as soon as possible and open issues for "Need to validate input", " Error handling in the case the analysis fails" and "The analysis in romanesco should use a separate Girder auth token, currently is uses the web clients token."   That way extending the contour task to leverage spark can happen in its own topic branch,  thoughts?

web_external/js/MinervaModel.js
```diff
@@ -9,6 +9,20 @@ minerva.models.MinervaModel = girder.models.ItemModel.extend({
...
+        // but to do that we would have to be persisting geoFileReader to the server
+        // which would require some things being rearranged.
+
+        // For now we know that if original_type is 'json' its ACTUALLY contour json,
+        // and if its'geojson'  its actually geojson - and for now these are the only
+        // two renderable data types.
+        return this.getMinervaMetadata().original_type === 'json' ||
+            this.getMinervaMetadata().original_type === 'geojson';
+    },
+
```
cjh1: Can you example the chose of name?

web_external/js/views/body/MapPanel.js
```diff
@@ -16,22 +16,26 @@ minerva.views.MapPanel = minerva.View.extend({
...
-        if (!_.contains(this.datasets, datasetId)) {
-            dataset.once('m:geoJsonDataLoaded', function () {
-                var layer,
-                    reader;
-                layer = this.map.createLayer('feature');
+        if (!_.contains(this.datasets, dataset.id)) {
+
+            dataset.once('m:dataLoaded', function (datasetId) {
+                var dataset = this.collection.get(datasetId);
+//              var minervaMeta  = dataset.getMinervaMetadata();
```
cjh1: Can we remove this?

web_external/js/views/body/AnalysisPanel.js
```diff
@@ -74,7 +77,7 @@
                 console.log('AP remove');
                 this.render();
             }, this);
-            this.render();
+            // this.render();
```
mgrauer: remove

web_external/js/views/body/DataPanel.js
```diff
@@ -101,7 +101,15 @@ minerva.views.DataPanel = minerva.View.extend({
...
-        this.render();
+
+        girder.eventStream.on('g:event.job_status', _.bind(function (event) {
+            var status = window.parseInt(event.data.status);
+            if (status === girder.jobs_JobStatus.SUCCESS) {
+                this.collection.fetch({}, true);
+            }
+        }, this));
+
+//        this.render();
```
mgrauer: remove

web_external/js/views/body/LayersPanel.js
```diff
@@ -45,7 +45,7 @@ minerva.views.LayersPanel = minerva.View.extend({
             girder.events.trigger('m:layerDatasetRemoved', dataset);
             this.render();
         }, this);
-        this.render();
+//        this.render();
```
mgrauer: remove

mgrauer:
Here's an example shapefile (you'll actually need 3 specific files all with the name shapefile, just use all 5) https://github.com/OpenGeoscience/minerva/blob/master/plugin_tests/data/shapefile.shp.  What I mean is, if you are using the local file uploader for a new dataset, select all 5 files named shapefile at one time.

mgrauer:
@kotfic can you also open issues related to splitting up the Dataset into specific subclasses?  It may end up being several issues, but here is what I want to have supported once all the issues have been resolved.  This is all current functionality.

  * have a dataset be able to differentiate whether it is contour json or json that needs to be mapped to geojson (could be based on the geo renderer type)
  * have a dataset based on a shapefile, which is then converted to a geojson file, be able to render the geojson file
  * have a dataset based on a json file, once a mapping is determined for lat/long, have the json be converted into geojson with points, be able to render the geojson file
  * have a dataset based on an external mongo collection (one that isn't mediated by Girder),  once a mapping is determined for lat/long, generate geojson with points, be able to render the geojson file
  * have a dataset that is a geojson file that can be rendered

I'm OK with not having support for CSV files for now, as the support for them up so far wasn't fully fleshed out.  We can easily provide an implementation of this similar to json/mongo once we get those done, or decide on some refinement such as only reading selected columns at the stage of uploading.

kotfic:
@mgrauer @cjh1 last commit fixes up comment issues, allowing shapefiles to be rendered etc  opened #62 for refactoring DatasetModel.js

mgrauer:
@kotfic gave me a demo of this, given that we've pushed additional work out to other issues, LGTM.

PR #60, opened 2015-07-30
Fix local file uploads

kotfic:
@mgrauer  dataFolderId was renamed on collection but not replaced - no instances of dataFolderId are left in minerva codebase


mgrauer:
Thanks for the fix @kotfic .  LGTM.

Issue #40, opened 2015-07-20
cache travis dependencies


mgrauer:
info from [this comment](https://github.com/OpenGeoscience/minerva/pull/36#issuecomment-122595593)

>  There are a few options to handle it:
  * Do what I did for romanesco and serve the dependencies as a tarball at data.kitware.com.
  * Use something like mason which does essentially the same thing but better. This will probably require a minor change to push builds to girder rather than s3.
  * Use circleci, which allows you persist directories between builds. I have used it as part of cdatweb testing to cache a very large docker image, and it works pretty well.

mgrauer:
We ended up caching pip and npm dependencies.

PR #61, opened 2015-07-30
WIP

mgrauer:


mgrauer:
@cjh1 @kotfic @jbeezley 

I just cleaned up this branch and force pushed.  I've added Travis caching for pip and Girder's node_modules dir.

At the start of this testing, I saw build times of 

    Push: 13:27
    PR: 14:10

At the end I saw

    Push: 3:53
    PR: 6:08

That last PR (6:08, which is gone b/c of my force push) failed to upload the cache in 180 seconds at the end--I'm not sure if that is included in the build time, my guess is that it is not.

We may be able to tune this a bit further with caching the global npm install, the install of R, and tuning the caching a bit, but the first order of magnitude work is done.  

Unless this introduces anything that scares anyone, I'd say this is an improvement towards #40 .

jbeezley:
Oh wow, I didn't know that travis allowed caching like that on the free accounts.  Anyway, LGTM.

mgrauer:
Only on containerized builds for public projects.  And maybe that is new?

cjh1:
I think you can do just:

cache: pip

Rather explicitly specifying the directory?

cjh1:
I guess not if you have overridden the install step ...

mgrauer:
@cjh1 yes, not if you override install step http://docs.travis-ci.com/user/caching/#pip-cache

I may just need to cache the pip cache dir and not site-packages, I'll explore that on some other PR.

PR #67, opened 2015-08-05
Removing romanesco's R dependency

cjh1:


cjh1:
@mgrauer This removes our R dependency ... using the new romanesco plugin system.

mgrauer:
I'm not sure if this is what you mean, but you can go to a Travis build, then go to settings menu/caches, and delete a cache from Travis.  My understanding is that it will clear out the cached files as they currently stand, and then re-cache the requested folders after the next run.

cjh1:
Yep, that is what I was trying todo by removing the cache statements, I am guessing it did a similar thing ...

mgrauer:
Build times are super fast now.  LGTM.

PR #73, opened 2015-08-12
Add jobs panel to session page

mgrauer:


mgrauer:
@kotfic PTAL.

This branch adds in a jobs panel placeholder.

It also fixes the Travis build, which was breaking because we weren't correctly installing geojs.  Now I've changed package.json to install geojs from npm.  We still have our instructions in the README that describe how to install any particular branch of geojs from Github.

kotfic:
@mgrauer I am getting the following error

```Uncaught TypeError: girder.views.jobs_JobListWidget is not a function```

I have updated to the current tip of master on girder,  and switched to the add_jobs_panel branch on minerva -  what did i miss?

mgrauer:
I put a commit on top of the build commit since I was hacking.  I'll close this PR out and make something a bit clearer.  Sorry for the confusion.

PR #74, opened 2015-08-13
Install geojs from npm to fix travis

mgrauer:
This fixes the Travis build, which was breaking because we weren't correctly installing geojs. Now I've changed package.json to install geojs from npm. We still have our instructions in the README that describe how to install any particular branch of geojs from Github.

@kotfic This should make more sense now.  :grinning:


jbeezley:
I'm glad it works.  This was the primary reason I made the release.

kotfic:
@mgrauer LGTM,  I am going to merge this so i can make sure my builds are working

Issue #65, opened 2015-08-04
Analysis should use seperate Girder auth token from webclient


kotfic:
currently contour analysis uses the webclient auth token,  but for long running jobs this token may be invalidated.

matthewma7:
Seems outdated.

Issue #39, opened 2015-07-20
Create a deployment/installation script for available analyses


mgrauer:
Possibly along the ansible path or the pip path.

From #29.

mgrauer:
To be a bit more specific here, there is Minerva installation, which includes running Girder, installing all the Minerva dependencies and running Minerva as the top level web app.  

Then there are things that will need to be included in Minerva's Girder instance to make the system workable, such as items with specific metadata that tie into Romanesco analyses.  This latter group is probably what is being referred to in #29.

mgrauer:
Perhaps these will be a configurable set of analyses depending on whether the user wants spark type analyses or bsve type.

mgrauer:
A possible path here is to have a particular folder where analyses live, and have a single analysis per module, with some module level properties describing the analysis.  Girder could then construct the analyses in the analysis folder (which are items containing specific metadata) from the modules in this analysis folder, including adding new analyses when new modules are detected.

This could be called from the ansible deployment script--we would need to decide on a Girder user to create these items as, this may just be the Girder admin user.

This could also be called from the module's load method, but then there is a bigger problem of figuring out which user to create these items as.

mgrauer:
As part of this deployment/installation process, the permissions for analysesXfolderXuser need to be specified in the Girder Romanesco plugin config.

PR #76, opened 2015-08-13
Mean contour analysis

kotfic:
This implements a contour analysis that takes the mean across time stamps for a particular variable.  It leverages the romanesco-spark integration and so expects a reachable spark cluster to be available. 

Please note to test this you will have to run the import_analysis.py script from a command line,  something to the effect of: 

```
$> cd /path/to/minerva/utility
$> python import_analyses.py \
--username kotfic \
--password letmein \
--port 8080 \
--path /path/to/minerva/analyses/NEX/
```


kotfic:
@mgrauer @aashish24 PTAL

aashish24:
:clap: Will try to run it on my machine.

kotfic:
@aashish24 I recommend using the following s3 bucket:
- s3 bucket name:  nasanex
- prefix:  /CMIP5/CommonGrid/gfdl-esm2g/rcp45/mon/r1i1p1/pr/
- file: regridded_1deg_pr_amon_gfdl-esm2g_rcp45_r1i1p1_200601-210012.nc

it takes about 30 seconds to download and about 30 seconds to process on my "cluster"  of four docker containers each with 2 cores

mgrauer:
@kotfic 

Testing this.  I have a spark master and slave both running on my host where Girder is running.  Do I need to do anything to tie Girder <=> Romanesco <=> Spark?

I have followed what you said about the S3 dataset.  'nasanex' is now coming up as an available dataset in the mean contour analysis UI.  What value do I enter in the parameter field?

kotfic:
@mgrauer  As long as spark is running at http://master:7077  then i believe it should "just work"   Otherwise I believe you can add configuration to the romanesco task like [this](https://github.com/Kitware/romanesco/blob/master/tests/spark_test.py#L14-L22)  (replacing os.environ['SPARK_TEST_MASTER_URL'] with the url of the spark head node). 

The paramater is 'pr'  (no quotes)

mgrauer:
@kotfic I'm getting `403 Forbidden {"message": "Unauthorized user.", "type": "access"}` when I try to run the analysis.  I'll ping you so we can set up a time to debug together.

mgrauer:
The 403 came out of Romanesco.  Due to the complexity of the reproduction environment for this issue, I will require some instructions for setting up Romanesco + Spark within Girder to be written before I sign off on this.  I can help with this, since I'm going through it now.

Including at least


  * install Romanesco via git
  * install Romanesco dependencies via pip
  * install Spark
  * run the Spark master
  * run the Spark slave
  * configure Romanesco 's worker.config.local, like so
```
[celery]
app_main=romanesco
broker=mongodb://localhost/romanesco

[romanesco]
# Root dir where temp files for jobs will be written
tmp_root=tmp
# Comma-separated list of plugins to enable
plugins_enabled=spark
# Colon-separated list of additional plugin loading paths
plugin_load_path=
```
  * start the Romanesco worker with `python -m romanesco`
  * grunt at the top level of Girder
  * enable the Romanesco plugin in Girder and restart Girder
  * configure the Romanesco plugin in Girder to allow user or group or folder access

@kotfic can you provide guidance on the correct way to configure the Romanesco plugin in Girder so that our permissions will be set correctly?  I.e., should this be for a specific user, or tied to the analysis folder, etc?

mgrauer:
Additionally, we need to 

  * set SPARK_HOME in the env where we run the celery worker
  * create a /data dir for this analysis that the celery worker can write to

mgrauer:
Sorry for the high message rate, for anyone who is listening.  If you want less noise, feel free to turn off notifications and let me know who you are, I'll ping you specifically when it is safe to turn notifications back on.

mgrauer:
At this point, the PR is most of the way there.  Changes we will want to get it across the finish line

  * revise the last commit to set a temporary work dir
  * add instructions (and figure out if or where to add them) for specifying the params of the m_c_a
  * decide on whether to further specify Girder/Romanesco permissions
  * try to find an easy path to processing a smaller file (this test takes about 3 min in Travis)

mgrauer:
Also, change the token in MeanContourWidget to a token separate from the user's login session, so that if the user logs out, the token for the job will still be valid.

mgrauer:
@nicopresto  This is the PR where we are adding Romanesco/Spark support to Minerva.  This particular analysis is just a rough example, but it proves out the full use of infrastructure, and is tested in Spark by Travis.

kotfic:
For the token problem there is already an open issue: https://github.com/Kitware/minerva/issues/65 In terms of permissions I"m going to say that is an important part of a larger conversation about how tasks get loaded into Minerva/Girder - load task is fine for examples,  but I think we want something more sophisticated down the line. Seems like that conversation should happen here: https://github.com/Kitware/minerva/issues/39

Finally in terms of finding a smaller file - i'm going to punt on this one.  Most of the file are quite a bit bigger and finding a smaller file doesn't mean it will be compatible with the script.

mgrauer:
Awesome, I'm glad this is finally going in, and tested!

:fireworks:

PR #77, opened 2015-08-14
Fixes hierarchy widget selectedItem bug

kotfic:
This bug is reproduced by opening a ReadOnlyHierarchyWidget, selecting
an item,  then clicking Use Selected.  Re-opening the same folder in the
widget shows the item as still being selected,  but clicking "Use
Selected" again causes the item to become unselected.

This is because the itemListWidget.checked variable in the
hierarchyWidget is maintaining the same 'checked'  state as
DatasetModel.attributes.meta.minerva.selectedItems.  the selectedItems
variable is used to update the checked boxes on render making it appear
that the box is checked,  when in fact it relies on the
itemListWidget.checked variable.  We will need to clean up the multiple
states so that only one variable tracks this information.

### DIFF
--- a/web_external/js/views/widgets/DatasetHierarchyWidget.js
+++ b/web_external/js/views/widgets/DatasetHierarchyWidget.js
@@ -28,6 +28,12 @@ minerva.views.DatasetHierarchyWidget = minerva.View.extend({
     },
 
     updateModelWithSelectedItems: function () {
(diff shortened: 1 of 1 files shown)


kotfic:
@mgrauer PTAL thanks!

mgrauer:
LGTM.  merging.

PR #85, opened 2015-09-02
Fix Spark caching in Travis

mgrauer:


mgrauer:
@kotfic PTAL.

The failures were due to caching issues.  In a now deleted commit as part of #76 , I had added the `worker.local.cfg` file, then removed it, thinking that doing a `pip` install of Romanesco would do this.  The tests continued to pass because that file was configured correctly in the cache.

To fix this, I've added creation and configuration of that file to our `.travis.yml`.  I tested this PR by deleted both Travis caches (PR & Push) before pushing this last commit.  Then I reran both builds to see that it would work with caching.

aashish24:
LGTM :+1:

PR #86, opened 2015-09-02
Server side source model

mgrauer:


mgrauer:
Ready for review.  This also includes commits from #85 as that is currently awaiting review.

PR #87, opened 2015-09-02
Server side wms source

mgrauer:


mgrauer:
Ready for review.  All commits but last are from other PRs.

PR #88, opened 2015-09-03
Client side source model

mgrauer:


mgrauer:
Ready for review.  This actually includes on the server side:

  * creating a WMS source
  * creating a WMS dataset

And on the Client side:

  * WMS Source Model
  * WMS Dataset Model

but no View.

mgrauer:
ping @essamjoubori @mbertrand

web_external/js/models/WmsSourceModel.js
```diff
@@ -0,0 +1,23 @@
+minerva.models.WmsSourceModel = minerva.models.SourceModel.extend({
+
+    createSource: function (name, baseURL, projection) {
+        var params = {
+            name: name,
+            baseURL: baseURL,
+            projection: projection
```
mbertrand: It's possible for a WMS server (and its layers) to support more than one projection.  Though for displaying on top of common basemaps like OSM, Google, etc, web mercator (EPSSG:900913/EPSG:3857) is likely the only one that matters.

mgrauer:
Good catch @mbertrand .  After our discussion on the GetCapabilities, it made more sense to me to have the projection be specified on the WMS Dataset as part of the params, once you have the capabilities in the WMS Source to show which projections are supported in which layers.  I had removed the field from the server side WMS Source model, but missed the client.  I've pushed a fix.

mgrauer:
ping @aashish24

server/rest/dataset.py
```diff
@@ -530,3 +532,35 @@ def createTweetGeocodes(self, item, params):
         .param('id', 'The Item ID', paramType='path')
         .errorResponse('ID was invalid.')
         .errorResponse('Write permission denied on the Item.', 403))
+
+    @access.public
+    @loadmodel(map={'wmsSourceId': 'wmsSource'}, model='item',
+               level=AccessType.READ)
+    def createWmsDataset(self, wmsSource, params):
```
aashish24: Why not create a class wmsDataset that inherits from dataset? And then have all datasets register themselves (in dataset or some other kernel)? In that way we can breakdown the dataset.py into multiple classes.

web_external/js/models/SourceModel.js
```diff
@@ -0,0 +1,2 @@
+minerva.models.SourceModel = minerva.models.MinervaModel.extend({
```
aashish24: Why not have the API on the SourceModel such as createSource?

server/rest/dataset.py
mgrauer: @aashish24  This is an API endpoint definition, and is so far the only functionality of the wmsDataset on the server-side.  I'm not sure what it would gain us to do what you suggest.

I can see having most of the code in this method be in a common method or superclass, but beyond that I'm not sure.

Maybe you can post some specifics or create an experimental PR with some suggested changes?

web_external/js/models/SourceModel.js
mgrauer: I could try it that way.  The subclasses would need to pass up the params and the path.

What I'm worried about is having a single method in the Source Model that will need to change as we add new Source types, given the fact that we don't have front-end testing, I'd be worried about possible breakage.  My preference for now is to build out the Source Model classes a bit, and then once we have some front-end testing in place do a unifying refactor.

I could do as you suggest though, I'm really on the fence and could be swayed by a strong argument.

server/rest/dataset.py
aashish24: Currently we have one file dataset.py that creates various dataset. As we add more dataset types this will get bigger and bigger. My suggestion with creating multiple files / classes will make this file smaller.

web_external/js/models/SourceModel.js
aashish24: @mgrauer I agree but I would suggest let me push another brach for clean up and all. I wanted to get the paver working as well (which let you setup your own dev environment). In the next branch I will do all the clean you need. For now, I wanted the vagrant route to work. If you don't want to merge this into master that's fine since this does not branch any dev-code changes.

mgrauer:
@aashish24 I have broken out the wms_dataset endpoint.

web_external/js/models/SourceModel.js
aashish24: Sure, we can revisit this later..

aashish24:
@mgrauer thanks for incorporating the suggestion. It would be nice if we further split the dataset into smaller modules.

aashish24:
Mostly looks good to me but let's wait for @mbertrand review.

mgrauer:
I agree with the further splitting, we can do that as part of the refactor.

aashish24:
Cool!.. do you want to wait for @mbertrand's review? if @mbertrand is happy then it looks good to me.

mgrauer:
I've got one more slight restructuring I want to do tomorrow.  I'll wait for @mbertrand , and I'm happy to talk it through with @mbertrand to help him get up to speed with Minerva.

web_external/js/collections/SourceCollection.js
```diff
@@ -0,0 +1,6 @@
+minerva.collections.SourceCollection = minerva.collections.MinervaCollection.extend({
+
+    model: minerva.models.SourceModel,
+    path: 'minerva_source'
+
+});
```
mbertrand: Should this conditionally set model to be minerva.models.WmsSourceModel (similar to the way DataSetCollection might return a WmsDatasetModel or S3DatasetModel depending on attrs.meta.minerva.dataset_type)?

server/rest/source.py
```diff
@@ -0,0 +1,115 @@
...
+        if folder is None:
+            raise RestException('User has no Minerva Source folder.')
+        name = params['name']
+        baseURL = params['baseURL']
+        desc = 'wms source for  %s' % name
+        item = self.model('item').createItem(name, user, folder, desc)
+        minerva_metadata = {
+            'source_id': item['_id'],
+            'source_type': 'wms',
+            'wms_params': {
```
mbertrand: Is 'wms_params' where username and password for authentication could be passed along?

mbertrand:
@aashish24 and @mgrauer, added a couple questions/comments this morning but overall this looks good to me.

server/rest/source.py
mgrauer: @mbertrand 

exactly right on 'wms_params'.

web_external/js/collections/SourceCollection.js
mgrauer: For sure, I'll push a fix.

mgrauer:
@mbertrand PTAL.

I did a bit more restructuring, splitting out wms_source to its own module and have addressed your comment about the Source Collection.  I'm satisfied with this branch if you are.

mbertrand:
@mgrauer I'm happy with the branch,  thanks!

PR #89, opened 2015-09-06
Ansible script to download and setup geonode, geoserver, postgres etc. 

aashish24:


aashish24:
* TODO: 
** Update the documentation
** Fix running geonode app 
** Fix creation of superuser

aashish24:
@mgrauer this is ready for review. Try this: 

> vagrant up production

and then 

> vagrant ssh production
> django-admin.py createsuperuser --settings=geonode.settings

then login to geonode:
http://localhost:8000

aashish24:
> vagrant ssh production
> source venvs/geonode/bin/activate
> django-admin.py createsuperuser --settings=geonode.settings

mgrauer:
I'm running into problems, but I'll try to push through them to get as far as I can.

If someone can QA #85 and it gets merged to master--the change is pretty simple and only affects travis--we can merge master into this branch and the tests should pass.

The source repo used as a basis for this has a highly irregular ansible structure, compared to what I've seen.  There is only a single role (and that role includes the original author's github handle), which then performs includes into other tasks, this makes debugging and maintenance tough.  We should probably restructure this to a more standard form.

mgrauer:
TASK: [ortelius.geonode | setup postgres cluster to default to utf8] **********
failed: [localhost] => {"failed": true}
msg: this module requires key=value arguments (['pg_dropcluster', '--stop', '9.4', 'main', '&&', 'pg_createcluster', '--start', '-e', 'UTF-8', '9.4', 'main', '#USE_SHELL'])

FATAL: all hosts have already failed -- aborting

mgrauer:
Sorry, that last error is with my ansible setup.  I've fixed that error by correcting my ansible setup.  For the record, I'm using ansible 1.9.3.

aashish24:
@mgrauer what was the error you ran into? I believe I am using 1.9.* as well. Also, do you think this branch is good to be merged into master?

mgrauer:
The errors I had were actually with my ansible install, not to do with this branch's code.

I've restructured things some in #90.  Let's merge master in there to see that the tests pass, and delete everything we don't need, clean up the documentation a bit, then we can merge.

aashish24:
@mgrauer what did you restructure?

PR #90, opened 2015-09-07
[WIP] Data services restructure

mgrauer:
@aashish24 

This is the start of restructuring the data_services setup.

There is still a lot of cruft from the original repo we'll want to remove for clarity.


mgrauer:
After I had done this, I was able to see a geoserver at http://localhost:8080/geoserver/web, but didn't test the setup any further than that (and don't know how to).

mgrauer:
I'm in the middle of things right now, so it is a bit messy and probably hard to see.  I'll need to delete the existing ansible_geonode dir once I'm finished.

The major restructuring I'm making is moving all of the tasks into separate roles.  Beyond that, I'm deleting all the stuff that we don't use.

aashish24:
@mgrauer can you make changes so that it show as diff and not new code. From the git point of view its looks odd that I have 20 commits and one commit brings all new changes. Also, as a developer that doesn't make me feel very excited :smile:. Also, please don't remove all the things we don't need right now since I was planning to make some more changes  and wanted the old code for reference. But if you don't like that idea, I can just work on my local copy.

mgrauer:
Good points, I got a bit wild in my restructuring.  I'll revise this so that it is much easier to follow.  I'll ping you on the new commits to let you see it evolve in a more sensible manner.

There was a lot of stuff in the old repo, including many docs, that made it difficult for me to see what was actually used and how to do things.  My goal was to make this all easier to read through and understand.

I don't think we should merge into master anything that we don't actually use.

jbeezley:
:+1: on keeping the file tree clean of unnecessary files.

PR #91, opened 2015-09-08
[WIP] Data services sane restructure

mgrauer:
This will allow much better visibility into restructuring the data services provisioning.


aashish24:
@mgrauer also if you vagrant box is running you can visit localhost;8000 and you should see geonode interface. If you had created superuser (like I suggested) you can use that to login into geonode and then upload a new layer by using some shapefiles from here: (http://www.statsilk.com/maps/download-free-shapefile-maps). Rest should be obvious. 

- Aashish

mgrauer:
I followed these [instructions](https://github.com/Kitware/minerva/pull/89#issuecomment-138192179).  Is there something specific I should enter for the username and password?  Is this username and password what I should enter on the geonode web UI, or should I login with other credentials there?

I'm rerunning this as is and will report back what I can see from the web UI.

aashish24:
Once you run 

> django-admin.py createsuperuser --settings=geonode.settings

and whatever username and password you create, you can then use that in geonode web UI. The default  username is admin and password is admin but Tomcat would not let you know use that unless you change tomcat settings.

mgrauer:
When I ran this as you said, it worked for me, nice!!

To pass QA, I would like

- [ ] the instructions you gave me in the PR comment to be documented in the codebase
- [ ] the ortelius.geonode role to be split out into common, uwsgi, geonode, nginx, postgres, geoserver tasks
- [ ] the playbook to be revised to call the separate roles instead of includes
- [ ] common role can just have apt-get update cache
- [ ] move defaults to group_vars/all
- [ ] remove AWS variables from group_vars/all
- [ ] remove AWS db creation code
- [ ] remove any reference to ortelius, except as we need for licensing
- [ ] remove any reference to riskinfo_lk
- [ ] remove any code that we don't actually use
- [ ] remove any documentation that is old or incorrect

@aashish24 I'll leave this to you to clean up, as you ask.  I'm happy to help here, but I'll wait to hear from you before doing anything.

aashish24:
> the ortelius.geonode role to be split out into common, uwsgi, geonode, nginx, postgres, geoserver tasks

@mgrauer why this is necessary? 

> the playbook to be revised to call the separate roles instead of includes

Same here.

> move defaults to group_vars/all
>  remove AWS variables from group_vars/all
>  remove AWS db creation code

Sure. Will do. 

> remove any reference to ortelius, except as we need for licensing
>  remove any reference to riskinfo_lk
>  remove any code that we don't actually use
>  remove any documentation that is old or incorrect

Will do.

mgrauer:
@aashish24 Both of those two points are tied together.

Roles should be separated because that makes understanding and maintaining the Ansible setup so much easier.  It also enhances future flexibility, and gives us the ability to split out roles among different servers should we want to.

Debugging is also enhanced, so that the messages displayed as Ansible is running will be "nginx | install nginx" instead of "ortelius.geonode | install nginx", as we get now because the includes.  It is also difficult to follow the flow right now, because everything is in a single role 'geonode', but that role actually has 3 include files, and one of those include files has 3 or so other includes.  It is much easier to grasp when you can see a single playbook calling all of the roles.  A final reason is that if we will want to combine this with a playbook for standing up Minerva, we will have some shared Ansible code between the two, and it will be much easier to add the shared stuff if everything is separated by role.

I think now is a good time to do it because you are familiar with the codebase.  Feel free to punt on this and add an issue instead.

At the very least to get this branch in, I want the role renamed from ortelius.geonode to geonode.  There is no need for us to have a role name after someone's github handle in our repo.

PR #94, opened 2015-09-08
GeoNode project with geoprocessing code for ForecastIO air temperature

mbertrand:
This is a fork of the standard GeoNode project template (https://github.com/GeoNode/geonode-project) that contains code for processing ForecastIO 's hourly global air temperature data and NASA's Global Flood Monitoring System (GFMS) data.  Instructions for setting up are in the README.rst file.  A sample configuration file for running celery under supervisor is also included (geonodegp/celery.conf).

PR #98, opened 2015-09-09
Added ansible to minerva data services infrastructure

aashish24:


aashish24:
- [ *] the instructions you gave me in the PR comment to be documented in the codebase
- [ *] the ortelius.geonode role to be split out into common, uwsgi, geonode, nginx, postgres, geoserver tasks
- [* ] the playbook to be revised to call the separate roles instead of includes
- [* ] common role can just have apt-get update cache
- [* ] move defaults to group_vars/all
- [* ] remove AWS variables from group_vars/all
- [* ] remove AWS db creation code
- [* ] remove any reference to ortelius, except as we need for licensing
- [*] remove any reference to riskinfo_lk
- [*] remove any code that we don't actually use
- [*] remove any documentation that is old or incorrect

aashish24:
@mbertrand it would be nice if you can run and test this as well. There is one issue I am running into is screen capture in geonode preview mode.

mbertrand:
@aashish24 I will try it out.  By screen capture do you mean issues with layer thumbnails?

aashish24:
Yes, layer thumbnails. I am using master of geonode but I am not sure if that's the problem because one time it did work for me.

data_services/deploy/ansible/roles/postgres_postgis/tasks/main.yml
```diff
@@ -0,0 +1,51 @@
...
+
+- name: setup postgres cluster to default to utf8
+  sudo: yes
+  shell: "pg_dropcluster --stop 9.4 main && pg_createcluster --start -e UTF-8 9.4 main"
+
+- name: update postgres client configuration
+  template: src=pg_hba.conf dest=/etc/postgresql/9.4/main/pg_hba.conf
+  sudo: yes
+  notify:
+    - restart postgres_postgis
```
mbertrand: This caused an error for me:
```
ERROR: change handler (restart postgres_postgis) is not defined
```

To fix I had to create a new file 'data_services/deploy/ansible/roles/postgres_postgis/handlers/main.yaml' with the following content:
```
- name: restart postgres_postgis
  service: name=postgresql state=restarted
  sudo: yes
```

mbertrand:
@aashish24 the thumbnail problem is with the SITEURL setting, it is set to 127.0.0.1, and the thumbnail URL's are generated as a concatenation of SITEURL + MEDIA_URL.  When accessing the site at http://localhost:8000, the thumbnails are trying to be loaded from http://127.0.0.1/uploaded/thumbs/*.png

I am looking into a possible fix for this.

data_services/deploy/ansible/roles/postgres_postgis/tasks/main.yml
aashish24: ah..I had that file but it wasn't committed, thanks for spotting it. I pushed it upstream.

aashish24:
> @aashish24 the thumbnail problem is with the SITEURL setting, it is set to 127.0.0.1, and the thumbnail URL's are generated as a concatenation of SITEURL + MEDIA_URL. When accessing the site at http://localhost:8000, the thumbnails are trying to be loaded from http://127.0.0.1/uploaded/thumbs/*.png
> 
> I am looking into a possible fix for this.

Great.thanks.

data_services/deploy/ansible/Vagrantfile
```diff
@@ -0,0 +1,22 @@
+# -*- mode: ruby -*-
+# vi: set ft=ruby :
+Vagrant.configure("2") do |config|
+  config.vm.box = "trusty64server"
+  config.vm.box_url = "https://oss-binaries.phusionpassenger.com/vagrant/boxes/latest/ubuntu-14.04-amd64-vbox.box"
+  config.ssh.username = 'vagrant'
+  config.ssh.forward_agent = true
+
+  config.vm.define :production do |production|
+  	production.vm.network :public_network, :bridge => 'eth0', :auto_config => false
```
mbertrand: As part of possible fix for thumbnail issue, try using a private network IP:
```
config.vm.network "private_network", ip: "192.168.33.12"
```

and remove the following:

```
    production.vm.network :public_network, :bridge => 'eth0', :auto_config => false
    config.vm.network "forwarded_port", guest: 8080, host: 8080
    config.vm.network "forwarded_port", guest: 80, host: 8000
```

data_services/deploy/ansible/group_vars/all
```diff
@@ -0,0 +1,54 @@
...
+# server directory to put the code into that is pulled from the code
+# repository
+app_code_dir: /home/{{deploy_user}}
+
+# location of the requirements.txt flie in the code
+requirements_file: "{{app_code_dir}}/requirements.txt"
+
+# root domain name the server should run as. do not include a subdomain,
+# for example, txt2react.com
+site_url: "http://127.0.0.1/"
```
mbertrand: Change site_url to private network IP:
```
site_url: "http://192.168.33.12/"
```

data_services/deploy/ansible/roles/geonode/templates/local_settings.py.j2
```diff
@@ -0,0 +1,56 @@
...
+        'PORT' : '5432'
+    }
+}
+
+# OGC (WMS/WFS/WCS) Server Settings
+OGC_SERVER = {
+    'default' : {
+        'BACKEND' : 'geonode.geoserver',
+        'LOCATION' : 'http://localhost:8080/geoserver/',
+        'PUBLIC_LOCATION' : 'http://{{server_name}}/geoserver/',
```
mbertrand: Use SITEURL as part of PUBLIC_LOCATION:

```
'PUBLIC_LOCATION' : SITEURL + '/geoserver/',
```

data_services/deploy/ansible/Vagrantfile
aashish24: Thanks @mbertrand. How does it fixes the thumbnail problem? I will try this change but just curious.

mbertrand: @aashish,

Currently the site on the VM is reached by going to http://localhost:8000 <http://localhost:8000/>, but the thumbnail URL's are generated as SITE URL + MEDIA_URL + /"thumbs/<thumbnail_name>.png --> http://127.0.0.1/uploaded/thumbs/<thumbnail_name>.png <http://127.0.0.1/uploaded/thumbs/%3Cthumbnail_name%3E.png> - which cannot be accessed.

By using a private network IP and setting SITEURL to that same IP, this problem is avoided because you can access the site at for example http://192.168.33.12 <http://192.168.33.12/> and the thumbnail URL will be generated as http://192.168.33.12/uploaded/thumbs/<thumbnail_name>.png <http://192.168.33.12/uploaded/thumbs/%3Cthumbnail_name%3E.png>

aashish24: Thanks @mbertrand. Makes sense. Although I am bit surprised because at one point :8000 work but it may because of some other config changes I made at the time. I think private IP sounds good to me. On this note, should we host this at Epidemico (that would be my preference) or Kitware (specially for GUMBO as we develop new data layers).

aashish24:
@mbertrand your suggestion worked!. I pushed a change. Please approve. @mgrauer any other concern?

mbertrand:
@aashish, great, looks good to me.  I am still thinking together with Nico about the best approach to hosting (and the best place for the code that would minimize merge/sync issues with our own EpiGeo repo - i.e. maybe as its own private repo, shared with Kitware, instead of inside a folder of Minerva?).  This PR sets up a vanilla version of GeoNode without any of EpiGeo's geoprocessing code to process ForecastIO air temp data and other sources.  I made a PR for a stripped-down version of the EpiGeo code here: https://github.com/Kitware/minerva/pull/94 <https://github.com/Kitware/minerva/pull/94>, but this doesn't use Ansible.  I assume you want a dev server with this geoprocessing code running on it?  Would it be acceptable to use the EpiGeo dev server as long as it it kept in sync with the shared code base that we both contribute to?

aashish24:
> I assume you want a dev server with this geoprocessing code running on it?  Would it be acceptable to use the EpiGeo dev server as long as it it kept in sync with the shared code base that we both contribute to?

If I understand it correctly. Yes, my idea was that we both share the general infrastructure code and have plugins for specific datasets processing. If we are using our shared project funding (kitware and epidemico) then the code should be in public domain since that's what we had in the document (we can make exception if needed to but that should be discussed per dataset). Then this infrastructure should be up and running at your place possibly with more plugins that you and team developed on your time and the code for that does not need to be shared (kitware can make a request for a particular dataset source code but you can accept or reject that if it has been developed on some other project funding). So In summary: 

* Develop common infrastructure code that can be run with ansible, celery etc. 
* Develop open source dataset processing plugins if using shared project funding 
* Keep other plugins local if not developed as part of the shared project funding.

aashish24:
@mgrauer can we merge this now? Thanks @mbertrand for your help and review!

mgrauer:
I'm provisioning a vm now with the final version, I will let you know the result of that when it finishes.

A couple of notes:

  1. I accidentally was on ansible v 1.8.2, and some postgres_ext didn't work.  When I switched to v 1.9.3 it worked (at least so far).  You could add a note about minimum ansible version.

  2. Can we name the VM something better than Minerva-production?  Maybe Minerva-dataservices?  This is specified in the Vagrantfile.  It isn't that important and we could decide later.

aashish24:
> I accidentally was on ansible v 1.8.2, and some postgres_ext didn't work. When I switched to v 1.9.3 it worked (at least so far). You could add a note about minimum ansible version.

Sure..can add a note in the docuemntation if that's okay. 

> 
> Can we name the VM something better than Minerva-production? Maybe Minerva-dataservices? This is specified in the Vagrantfile. It isn't that important and we could decide later.

Hmm.. @mbertrand what you think? You decide. I am fine with anything.

mgrauer:
@aashish24 This is great, a really clean setup.  We can use this as the basis for our total Minerva stack deployment at some later point (where we get Minerva + data services both provisioned).

We can punt on the VM name for now, we'll revisit it with later ansible work.

I give everything a LGTM.  Since I pushed one final commit, can someone OK it once Jenkins finishes?

aashish24:
Great... LGTM

mbertrand:
@aashish I'm ok with any VM name, though I guess I like '-dataservices' more than '-production'.

aashish24:
Sure.I will push a branch to rename it to minerva-dataservices

PR #108, opened 2015-09-15

essamjoubori:


aashish24:
Nice @essamjoubori :clap:.. ..I will review it after @mgrauer does it.

mgrauer:
@essamjoubori I'll start listing QA comments here, though I won't finish this today.

-  [ ] change the name of the branch from `essam` to something more meaningful; if you are on the head of this branch on your command line, you can do `git checkout -b wms_display` or similar, then you'll have to push it and create a new PR.  Do you have the permissions to push directly to this repo?
-  [ ] there is already a [`.jshintrc`](https://github.com/Kitware/minerva/blob/a34b9abb6561fd07a1a5ecad4411d3dded592752/plugin.cmake#L51) that we merge with the overall Girder one, you can make additions there if need be.  Can you remove the one you added at the top level?
- [ ] remove the three files added under `src`, add `src` to the .gitignore (I should have done this before now :grin:: )

essamjoubori:
@mgrauer I've made the changes and created a new PR with the same title. You may close this one.

I don't believe I have the permissions to push directly to this repo.

Thanks!

jbeezley:
We should get the Epidemico folks permissions on this repo.  Currently all of us Kitware people have access with the Arbor workflows group.  I'm thinking we should create a new group for this purpose.  What do you think @aashish24?

aashish24:
Thanks for bringing this up @jbeezley. I will fix the group and the permissions.

aashish24:
@jbeezley should be all set. I added Matt and others. @essamjoubori let me know if I missed someone.

PR #110, opened 2015-09-17

essamjoubori:


server/rest/wms_dataset.py
```diff
@@ -34,12 +36,28 @@ def __init__(self):
     @loadmodel(map={'wmsSourceId': 'wmsSource'}, model='item',
                level=AccessType.READ)
     def createWmsDataset(self, wmsSource, params):
+        # Get layer legend (TODO// Include authentication in the future)
+        # Legend to be included in the metadata?
+        base_url = wmsSource['meta']['minerva']['wms_params']['base_url']
+        hostName = base_url[7:28]
```
mgrauer: call [getUrlParts](https://github.com/girder/girder/blob/master/girder/api/rest.py#L39-L57)

your current method is more brittle, and wouldn't give you what you expect for e.g. https

server/rest/wms_source.py
```diff
@@ -20,20 +20,33 @@
...
 
 class WmsSource(Source):
+
     def __init__(self):
         self.resourceName = 'minerva_source_wms'
         self.route('POST', (), self.createWmsSource)
 
     @access.user
     def createWmsSource(self, params):
+        print params
```
mgrauer: remove stray print

server/rest/wms_source.py
```diff
@@ -20,20 +20,33 @@
...
         name = params['name']
         baseURL = params['baseURL']
+        username = params['username'] if 'username' in params else None
+        password = params['password'] if 'password' in params else None
+        wms = WebMapService(baseURL, version='1.1.1',
+                            username=username,
+                            password=password
+                            )
+        layers = list(wms.contents)
         minerva_metadata = {
```
mgrauer: We will want to call this this map service asynchronously in the future, but that will add complexity to the UI, so let's just track it with an issue for now.

https://github.com/Kitware/minerva/issues/112

web_external/js/models/DatasetModel.js
```diff
@@ -6,6 +6,10 @@ minerva.models.DatasetModel = minerva.models.MinervaModel.extend({
         files: null
     },
 
+    isWmsSource: function () {
```
mgrauer: I don't like this, but I totally understand why you did it, and it makes sense given that we don't have clear separation in the UI b/w sources and datasets.

I'll add an issue to refactor this once the source list appears in the UI.

https://github.com/Kitware/minerva/issues/113

web_external/js/views/widgets/AddWmsSourceWidget.js
```diff
@@ -0,0 +1,35 @@
+'use strict';
```
mgrauer: remove this use strict

web_external/js/views/widgets/LegendWidget.js
```diff
@@ -0,0 +1,29 @@
+'use strict';
```
mgrauer: remove this use strict

web_external/js/views/widgets/WmsLayersListWidget.js
```diff
@@ -0,0 +1,54 @@
+'use strict';
```
mgrauer: remove this use strict

This is probably very hard to see, but you don't need to include `use 'strict';`.  It will be done automatically by the uglification process when all of the JS source in Minerva is combined into a single minerva.min.js .

I had to do a bit of digging on this myself to track it down.

web_external/js/views/widgets/WmsLayersListWidget.js
```diff
@@ -0,0 +1,54 @@
...
+/**
+* This widget displays the list of WMS layers
+*/
+minerva.views.WmsLayersListWidget = minerva.View.extend({
+
+    events: {
+        'submit #m-add-layers-form': function (e) {
+            e.preventDefault();
+            var wmsSource = this.dataset;
+            var hostName = (wmsSource.get('meta').minerva.wms_params.base_url).slice(0, 28);
```
mgrauer: Since we need the separate parts of the base url, go ahead and parse it out on the server side with the getUrlParts method I mentioned, then save the parts into the metadata of the dataset.  Then we should be able to access them here.

web_external/js/views/widgets/WmsLayersListWidget.js
```diff
@@ -0,0 +1,54 @@
...
+            var that = this;
+            $('input[type=checkbox]').each(function () {
```
mgrauer: change this to `_.bind` for consistency (at least I hope we are being consistent).  That is, remove `var that = this;`, and

    $('input[type=checkbox]').each(_.bind(function () {
        // some stuff
    }, this));

 and then you can change the `that`s to `this`.

mgrauer:
@essamjoubori 

My apologies that this QA is taking so long, thank you for your patience.

Can you make the changes I've asked for here, and then create a new (yet again, sigh) PR by pushing this branch directly to Kitware/minerva?  This will exercise your permissions to this repo and also allow me to push some commits onto your branch more easily.

mgrauer:
This PR was mistakenly merged.  I've force-pushed master, and also set master as a protected branch.  This means we can't force push to master, and branches can only be merged to master if Travis tests pass.

ping @essamjoubori  @mbertrand @aashish24 @jbeezley @kotfic

aashish24:
@mgrauer how did that happen?

mgrauer:
The wrong button was accidentally pushed, merge instead of close.  This branch protection won't protect us from that situation, unless the PR isn't currently passing.  It will protect against force pushes.

PR #115, opened 2015-09-18

essamjoubori:


mgrauer:
@essamjoubori I would like to take over this branch for a bit.  I'll push some commits and then let you know when it is ready for you to look at or work on further.  I'm curious about the '8', you can let me know in a comment on the PR, since I'll be updating that MapPanel js file.

mgrauer:
I was getting some weird errors in Travis with 

>UNMET DEPENDENCY grunt@~0.4

I think it was a caching issue--the PR itself wasn't even listed among the PRs Travis had a cache for.  I deleted all the Travis caches and reran--now the branch still fails, but in a much more reasonable way, i.e. because tests actually fail.

server/rest/wms_dataset.py
```diff
@@ -34,15 +37,32 @@ def __init__(self):
     @loadmodel(map={'wmsSourceId': 'wmsSource'}, model='item',
                level=AccessType.READ)
     def createWmsDataset(self, wmsSource, params):
+        # Get layer legend (TODO// Include authentication in the future)
+        # Legend to be included in the metadata?
+        baseURL = wmsSource['meta']['minerva']['wms_params']['base_url']
+        parsedUrl = getUrlParts(baseURL)
+        hostName = parsedUrl.netloc
+        typeName = params['type_name']
```
mgrauer: Can you change this param to the endpoint from `type_name` to `typeName`?  I know we lack consistency in our styles, but let's be consistent within an individual endpoint.

Also, can you add `typeName` to the endpoint Description and make it a required param?

mgrauer:
@essamjoubori This was great stuff, thanks!  I think we are very close here.  

Can you make the changes I requested on the createWmsDataset endpoint?  Also the wms_test is failing, related to that same param.  Can you fix this one also?  The first error was that the test was missing the `type_name` (soon to be `typeName`) param, though there may be other errors.

jbeezley:
@mgrauer I saw that error earlier and removed the cache.  It worked once, then failed again after reusing the cached directory.  There must be some sort of dependency collision, but it isn't obvious to me where.  I may try a new PR that updates to iojs.  The node version 0.10 used here is starting to fall out of compatibility.

essamjoubori:
@mgrauer I've made the changes to the endpoint. I ran the tests locally and they all passed except for one related to this line: https://github.com/Kitware/girder/blob/master/tests/web_client_test.py#L160. However, I wasn't able to reproduce this issue on another machine. Also, Travic still not happy.

`ERROR: testWebClientSpec (tests.web_client_test.WebClientTestCase)
Traceback (most recent call last):
File "tests/web_client_test.py", line 160, in testWebClientSpec
     line = line.decode('utf8')
File "/home/vagrant/.virtualenvs/girder/lib/python2.7/encodings/utf_8.py", line 16, in decode
     return codecs.utf_8_decode(input, errors, True)
UnicodeDecodeError: 'utf8' codec can't decode byte 0xcc in position 122: invalid continuation byte`

mgrauer:
Thanks @essamjoubori .

I'm looking into Travis now.

mgrauer:
@essamjoubori 

I just pushed my last commit onto this PR, I'll give it back to you now.  I think Travis was fixed by no longer caching any of the npm packages, we'll have to keep an eye on it.

We have one last issue to fix on this PR, @mbertrand said you were working on it now, related to the assumption of layer names being prefixed with "geonode:".

mgrauer:
@essamjoubori  This last commit fixes WMS for me, I can display geonode or topps sources now.  I give this a LGTM.  I'll wait for you to QA my commits.  If they look good to you, you can merge.

essamjoubori:
LGTM! Thanks @mgrauer

aashish24:
:clap: Nice @essamjoubori

Issue #113, opened 2015-09-17
Refactor isWmsSource in DatasetModel once SourceListView is created


mgrauer:
most likely the thing to do is remove `isWmsSource`, and have something specific to WmsSource and WmsDataset render correctly in the appropriate list, Source or Dataset respectively.

For now this function is necessary because Sources and Datasets are listed in the same view.

mgrauer:
additionally the showWmsLayersList method should probably be moved from the DataPanel to the SourcePanel

mgrauer:
and the section in AddDataSetWidget like

    } else if (dataSetType === 'm-mws-source') {

but obviously the other sources will have to be pulled out of there as well

Issue #95, opened 2015-09-09
createWmsDataset endpoint should be access.user instead of access.public


mgrauer:
https://github.com/Kitware/minerva/blob/03ba2065686892d0676aba5b41504d74e80a8178/server/rest/wms_dataset.py#L33

PR #117, opened 2015-09-19
Add source panel

mgrauer:
Based on #115.  Will need to be rebased once that goes into master.

### DIFF
--- a/plugin_tests/source_test.py
+++ b/plugin_tests/source_test.py
@@ -72,7 +72,7 @@ class SourceTestCase(base.TestCase):
         }
         response = self.request(path=path, method='GET', params=params)
         self.assertStatusOk(response)
-        folder = response.json
+        folder = response.json['folder']
         self.assertEquals(folder, None)
 
         # create a source folder
@@ -83,7 +83,7 @@ class SourceTestCase(base.TestCase):
         response = self.request(path=path, method='POST', params=params, user=self._user)
 
         self.assertStatusOk(response)
-        folder = response.json
+        folder = response.json['folder']
         self.assertNotEquals(folder, None)
         self.assertEquals(folder['baseParentType'], 'user')
         self.assertEquals(folder['baseParentId'], str(self._user['_id']))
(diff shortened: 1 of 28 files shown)


mgrauer:
Fixes #113.  Fixes #95.

Still an issue with WMS layers being added multiple times.  Add them from a source, delete them.  Add them again, 2 are added, delete them.  Add them again, 3 are added.  etc.

Removing the line
  
    this.$el.children().detach();

may have been premature, but putting it back in doesn't seem to fix the issue.  I'll look into it further tomorrow.  

@essamjoubori, let me know if you have a fix for this.

essamjoubori:
@mgrauer I faced this issue before. I believe this is what's happening: the `showWmsLayersList` function (in views/DataPanel) is called every time you show the list of layers from the source and that will create multiple `WmsLayersListWidget` views and duplicated events.
I solved this issue before by adding this conditional statement: https://github.com/Kitware/minerva/blob/wms_display/web_external/js/views/body/DataPanel.js#L143-Lundefined
I wasn't able to reproduce the bug. Let me know if you still having this issue.

mgrauer:
Thanks for the clue @essamjoubori .  I changed that code you referred to because of a bug, but my change introduced a new bug.  At this point they are both fixed.

The only problem I see left on this branch is an error where you have a WMS Layer Dataset, you add it to the map and then remove it from the map, and then try to delete the dataset, you get an error message from GeoJs saying

>Uncaught TypeError: Cannot read property '_exit' of null

I found a fix for this while investigating and will push shortly.

mgrauer:
Rebased on current master now that #115 is merged.  

Ready for review @essamjoubori @mbertrand

mbertrand:
@mgrauer, @essamjoubori  I tried it out, looks good.  I'm experiencing an issue though when trying to re-open a saved session:  I get a mostly blank page under the header with a javascript exception being "Uncaught TypeError: Cannot read property 'createLayer' of undefined" in MapPanel.js (https://github.com/Kitware/minerva/blob/master/web_external/js/views/body/MapPanel.js#L59).  But this happens for me on the master branch too.

mgrauer:
Thanks for the report @mbertrand .  I'd like to fix that in a separate PR, as this one is long enough as is, and as you say, this also exists on master.

mgrauer:
I'm going to merge this based on @mbertrand 's LG.

PR #119, opened 2015-09-22
Load saved wms in session

### LINKED ISSUE #118: Session with saved WMS layer fails to display when loaded
From this [comment](https://github.com/Kitware/minerva/pull/117#issuecomment-142347494)

mgrauer:
fixes #118.

@mbertrand this is the bug you found.  I'm glad you mentioned it, that part of the code was kinda confusing, I think this is an improvement in clarity.

### DIFF
--- a/web_external/js/views/body/LayersPanel.js
+++ b/web_external/js/views/body/LayersPanel.js
@@ -10,42 +10,12 @@ minerva.views.LayersPanel = minerva.View.extend({
         dataset.set('displayed', false);
     },
 
-    addDatasetToLayers: function (dataset) {
-        girder.events.trigger('m:layerDatasetLoaded', dataset);
-    },
-
     initialize: function (settings) {
         settings = settings || {};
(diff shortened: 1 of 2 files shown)


mgrauer:
I also had to do a bit of Travis work on this branch.  Note that commit 66aa0a9 and before are all already in master.

mgrauer:
Rebased this PR so that it is much easier to QA.  

To see the problem this commit fixes: on master, create a WMS source, then a dataset layer from that source.  Add the layer to the map, then save the session.  Now refresh the page, and you'll see an error.

aashish24:
@mgrauer can you update this branch?

aashish24:
the changes looks reasonable but I didn't run the code.

mgrauer:
@mbertrand you reported this bug initially, would you mind checking if it fixes your issue?

mbertrand:
@mgrauer it does fix the issue, thanks

PR #120, opened 2015-09-23
Update cmake binary URL

jbeezley:
Also, setting `-L` to follow in future redirects.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -59,7 +59,7 @@ before_install:
 
     # cmake
     - cd $HOME
-    - curl "http://www.cmake.org/files/v3.1/cmake-3.1.0-Linux-x86_64.tar.gz" | gunzip -c | tar x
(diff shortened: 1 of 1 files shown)


mgrauer:
@jbeezley I did similar work in aeec9827efe2f9aae8eb54efd7e8c17c30528ecf .

I didn't add -L in mine, though, this is a good addition.  

I did split out the piping operations so that I could do a travis_retry specifically on the curl step--according to the docs this will retry a network operation 3 times, and it happens automatically for any commands in the `install` block which is why I moved some npm commands into that block--my thinking here was that it'll retry only on a failure of the curl operation and not the subsequent unpacking.

I'll give you the option, you can include any of those retries here, or this LGTM as is, I don't think it matters that much.  Either way I'll clean up my branch after you merge to master.

jbeezley:
I'll go ahead an merge it for now because builds are failing without it.

PR #123, opened 2015-09-24
Don't use the OSM attribution when using a WMS source

jbeezley:
This should get rid of the duplicate attributions that you were seeing.


aashish24:
LGTM :+1:

PR #126, opened 2015-09-28
Add fedora install instructions

danlamanna:


mgrauer:
@danlamanna Is this done?  LGTM if so.

Issue #121, opened 2015-09-23
404 query params error when creating WMS layer


mgrauer:
I've seen this as a fleeting error a few times, it always seems to resolve itself as soon as I try again, and I haven't got the details down yet.  

Sometimes when a WMS source exists and you try to create a new layer from that source, there is a 404 query params error.  I've seen it with Flood Detection/Intensity, but that may just be because it is what I was playing with recently.

@essamjoubori  @mbertrand Please be on the lookout and if you see it provide more details.

:eyes:

mgrauer:
http://localhost:8080/?Proportion+of+Outpatient+Malaria+Visits=on#session/5602eaaf0640fd0634293a71

Unexpected query string parameters: Proportion of Outpatient Malaria Visits


Failed to load resource: the server responded with a status of 405 (Method Not Allowed)
http://localhost:8080/?Proportion+of+Outpatient+Malaria+Visits=on#session/5602eaaf0640fd0634293a71 Failed to load resource: the server responded with a status of 404 (Not Found)

mbertrand:
@mgrauer I noticed this too today, will look into it

PR #127, opened 2015-09-28
Fix for issue 121

mbertrand:
@mgrauer This should fix the issue you reported (https://github.com/Kitware/minerva/issues/121)

Steps to reproduce issue:
- Add a new WMS source
- Add a dataset from the source
- Add another WMS source
- Add a dataset from the second WMS source

### DIFF
--- a/web_external/js/views/body/SourcePanel.js
+++ b/web_external/js/views/body/SourcePanel.js
@@ -20,17 +20,13 @@ minerva.views.SourcePanel = minerva.View.extend({
     displayWmsLayersList: function (evt) {
         var el = $(evt.currentTarget);
         var wmsSource = this.sourceCollection.get(el.attr('cid'));
-        if (!this.wmsLayersListWidget) {
-            this.wmsLayersListWidget = new minerva.views.WmsLayersListWidget({
-                el: $('#g-dialog-container'),
-                source: wmsSource,
-                collection: this.datasetCollection,
-                parentView: this
-            });
-            this.wmsLayersListWidget.render();
-        } else {
-            this.wmsLayersListWidget.setCurrentSource(wmsSource);
-        }
+        this.wmsLayersListWidget = new minerva.views.WmsLayersListWidget({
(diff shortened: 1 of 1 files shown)


mgrauer:
Thanks @mbertrand .  This fixes the 404, but if you add a layer from a wms source, then add a layer from that same source, it will add multiple layers (2 the second time, 3 the third time).  Every time you create a new widget, you attach another event listener to the form, which is why we have the multiple adds.  

Your fix pointed me in the right direction, though, I've got a fix pushed.

web_external/js/views/body/SourcePanel.js
```diff
@@ -20,17 +20,13 @@ minerva.views.SourcePanel = minerva.View.extend({
...
-                el: $('#g-dialog-container'),
-                source: wmsSource,
-                collection: this.datasetCollection,
-                parentView: this
-            });
-            this.wmsLayersListWidget.render();
-        } else {
-            this.wmsLayersListWidget.setCurrentSource(wmsSource);
-        }
+        this.wmsLayersListWidget = new minerva.views.WmsLayersListWidget({
```
mbertrand: @mgrauer, if I add this snippet just before line 23 then the dupe issue goes away:
```
        if (this.wmsLayersListWidget) {
            this.wmsLayersListWidget.destroy();
        }
```
But maybe this isn't as efficient as reusing the existing widget.  What do you think?

mgrauer:
That is good to know @mbertrand , and it makes sense, thanks.

I'd rather go with the solution in #128 to keep consistent with best practices in Girder, especially as I think using the `girderModal` will be needed to set up client side testing correctly (or at least so we can tap into the existing testing infrastructure and institutional knowledge) once we finally get to testing there.

mbertrand:
Thanks @mgrauer, should I go ahead and close this PR?

mgrauer:
Yes, I'll do it, just wanted to complete the discussion first.

PR #125, opened 2015-09-28
Handle WMS authentication with encrypted credentials and proxying

mbertrand:
If  a username & password are entered in for a WMS Source, they will be encrypted and saved in the metadata for the source and it's child datasets.  The credentials will then be decrypted, passed on to a proxy, and added to headers of GetMap requests.

A sample encryption key was added to server/constants.py but should be customized for a production instance.

PR #128, opened 2015-09-29
Correct usage of girderModal

### LINKED ISSUE #121: 404 query params error when creating WMS layer

mgrauer:
Fixes #121.

Steps to reproduce issue:
- Add a new WMS source
- Add a dataset from the source
- Add another WMS source
- Add a dataset from the second WMS source

Also check that adding a dataset from a source, then adding the same or different dataset from the same source doesn't add multiple/duplicate datasets.

It was my fault before @mbertrand , I wasn't using the right magic to invoke modal dialogs.  We probably need to ensure we are doing this everywhere.

### DIFF
--- a/web_external/js/views/widgets/WmsLayersListWidget.js
+++ b/web_external/js/views/widgets/WmsLayersListWidget.js
@@ -37,9 +37,10 @@ minerva.views.WmsLayersListWidget = minerva.View.extend({
     },
 
     render: function () {
-        var modal = this.$el.html(minerva.templates.wmsLayersListWidget({ layers: this.layers }));
+        var modal = this.$el.html(minerva.templates.wmsLayersListWidget({
+            layers: this.layers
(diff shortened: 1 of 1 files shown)


mgrauer:
@mbertrand is this one OK by you?  I'll wait until #125 goes in before merging, so you won't have to rebase yet again :grin: .

mbertrand:
@mgrauer I tried it out and it works, looks good!

Issue #131, opened 2015-09-29
Move WMS encryption key into a girder configuration file.


mbertrand:
Current location of the WMS encryption key is in minerva/server/constants.py ("CRYPTO_KEY").

Move it to one of the girder configuration files (girder.dist.cfg or girder.local.cfg) and update the relevant code to find it there.

PR #132, opened 2015-09-30
Minerva configuration file for encryption key

mbertrand:
Created minerva configuration files (conf/minerva.dist.cfg, conf/minerva.local.cfg), moved the crypto_key setting out of constants.py into the minerva.dist.cfg file, and added code to read in the configuration.

Also removed some duplicate code from the WmsProxy class.

@mgrauer This PR addresses https://github.com/Kitware/minerva/issues/131

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,6 +31,7 @@ before_install:
     - rm -rf girder
     - git clone https://github.com/girder/girder.git
     - mv Kitware/minerva girder/plugins
+    - cp girder/plugins/minerva/server/conf/minerva.dist.cfg girder/plugins/minerva/server/conf/minerva.local.cfg
 
     # clone romanesco
     - cd $HOME/build/girder/plugins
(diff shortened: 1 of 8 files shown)


mgrauer:
Thanks @mbertrand, this LGTM.

I made two minor changes

    * gitignoring and removing minerva.local.cfg and explaining the local vs dist in readme
    * namespaced the config to minerva, to give it a bit of separation from the girder config

Do these seem ok to you?

I would like to test the encryption when creating a wms_source, but that would require us having a known endpoint that we can connect to from travis, with authentication.

mbertrand:
@mgrauer those changes look good to me.  I will create a test for the encryption.

mgrauer:
@mbertrand  I'm going to merge this PR.  Encryption testing can come in a separate PR.  That's probably what you intended anyway, rereading your comment.

PR #143, opened 2015-10-06
[WIP] Bsve to geojson

mgrauer:
I don't think we'll want to merge this, at least not in its present form.  It's an experiment with specifying the type of data we get back from bsve search and also allowing for a bit of control over how geojson points are rendered.

This is an improvement over the current UI, and will be good to demo and gather feedback.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,6 +31,7 @@ before_install:
     - rm -rf girder
     - git clone https://github.com/girder/girder.git
     - mv Kitware/minerva girder/plugins
+    - cp girder/plugins/minerva/server/conf/minerva.dist.cfg girder/plugins/minerva/server/conf/minerva.local.cfg
 
     # clone romanesco
     - cd $HOME/build/girder/plugins
--- a/README.md
+++ b/README.md
@@ -46,9 +46,10 @@ sudo dnf install git gcc-c++ libffi-devel make python-devel python-pip freetype-
 
     grunt
 
-- change the encrypt_key value in Minerva's `minerva.dist.cfg` file, located in the server/conf directory.
-The key needs to be a 32 byte url-safe base-64 encoded string.  You can either replace the existing string
-with one of equal length, using letters and numbers, and ending with an '=', or generate one within python
+- copy the `minerva.dist.cfg` file, located in the server/conf directory, to `minerva.local.cfg` in that same directory.
+Any property in `minerva.local.cfg` will take precedent over any property with the same name in `minerva.dist.cfg`.
+Change the `encrypt_key` value in `minerva.local.cfg` file; the value should be a 32 byte url-safe base-64 encoded string.
+You can either replace the existing string with one of equal length, using letters and numbers, and ending with an '=', or generate one within python
 with the following code::
 
 ```
--- a/README.md
+++ b/README.md
@@ -46,6 +46,16 @@ sudo dnf install git gcc-c++ libffi-devel make python-devel python-pip freetype-
 
     grunt
 
+- change the encrypt_key value in Minerva's `minerva.dist.cfg` file, located in the server/conf directory.
+The key needs to be a 32 byte url-safe base-64 encoded string.  You can either replace the existing string
+with one of equal length, using letters and numbers, and ending with an '=', or generate one within python
+with the following code::
+
+```
+from cryptography.fernet import Fernet
+Fernet.generate_key()
+```
+
 - enable the Minerva plugin through the Girder Admin console
 - restart Girder through the Girder Admin console
 
--- a/server/__init__.py
+++ b/server/__init__.py
@@ -16,8 +16,16 @@
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
 ###############################################################################
-
 import loader
+import os
+from girder.utility.config import _mergeConfig
+
+PACKAGE_DIR = os.path.dirname(os.path.abspath(__file__))
+
+# Read the configuration files
+_cfgs = ('minerva.dist.cfg', 'minerva.local.cfg')
+for f in _cfgs:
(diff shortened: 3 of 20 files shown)


mgrauer:
Let's say you do a search of the bsve api with this query

```
{
        "clientTimeZone": "EST",
        "fromDate": "2015-01-01",
        "locations": [
            {
                "location": "United States",
                "locationType": "country"
            },
            {
                "location": "USEUCOM",
                "locationType": "cocom"
            }
        ],
        "sources": [
            "ProMED",
            "AP Top Science News",
            "WebSearch"
        ],
        "term": "dengue",
        "toDate": "2015-10-02"
    }
```
and a similar one replacing `dengue` with `flu`.

These will be returned as JSON, which will be converted to a renderable Geojson dataset, all when you run the `bsve_search` analysis.

By default you can render the results as points.

screen shot 2015-10-06 at 3 52 08 pm


You can also click on the paintbrush icon next to the dataset

screen shot 2015-10-06 at 3 58 54 pm


and ask that the dataset be rendered with Geojs' cluster feature, and specify a color scheme.


screen shot 2015-10-06 at 3 56 47 pm


Here I've set a different color scheme for `flu` and `dengue`, then rendered them both at the same time.

screen shot 2015-10-06 at 3 57 18 pm

aashish24:
This is awesome @mgrauer specially the part of color scheme, that is altogether a new capability which will make a huge impact in the future. 

Do you want us to review it?

danlamanna:
@aashish24 I think the point clustering that @mgrauer implemented makes sense for us to use for Terra use case 1 (we discussed Friday).

aashish24:
@danlamanna agree +1, the clustering feature in GeojS is implemented by @jbeezley @mgrauer exposed it in Minerva.

mgrauer:
Hmmm, good question. 

Let's say, don't review it as if it would go into master.  You can take a look at the MapPanel.js additions to get a sense of how I did this, but these need to be improved.  Probably having someone with a better understanding of GeoJs would be helpful.  I'll improve it after I've got some feedback on the functionality and get a better sense of where want to go next.

My main sticking point right now is that the adapter between Geojson rendering in MapPanel and GeoJs feels awkward and too specific.  I can see this breaking out into its own JS file so we can have some more flexibility.

Probably more useful is a review of the UI/functionality, which you can give me after the demo.

mgrauer:
I should amend that, if you have some UI/functionality suggestions, let me know as it is possible I could make some quick improvements there.

mgrauer:
One addition I want to get in is a little color scheme depiction next to the session layer, i.e. you would be able to see that flu is the blue color scheme and dengue is the brown one by looking at the Session Layers panel.  Right now you can't tell which is which.

mgrauer:
Added a display of the cluster color scheme, in the session layer panel, should one be set.

screen shot 2015-10-07 at 6 31 49 pm

aashish24:
@mgrauer what's holding this branch?

mgrauer:
This isn't very general or well thought out, it was more of an experiment.  It provides support for colored clusters by littering the codebase in a few different places with code specific to this use case.  Possibly some of the bsve analysis stuff can be broken out and go in soon.

Here are some of the questions I have, maybe you @aashish24 and @jbeezley could provide suggestions, or @danlamanna could chime in since he has need of the cluster display.

  * I'm not saving any of the other data fields when I convert the bsve json to geojson.  Should I allow the user to remap and perform the conversion again?
  * specifying the rendering of geojson makes sense assuming all you have is points, but what should the UI look like if your geojson is a mixture of points and other objects?
  * similarly, we either render as geojson using the geojs reader, or else we assume points and create the points from features.  how can we render the other geojson objects other than the points if we have a geojson with a mix of features?  look specifically in the [MapPanel](https://github.com/Kitware/minerva/blob/bsve_to_geojson/web_external/js/views/body/MapPanel.js#L50-L123).  Is there some way I can render the points as is but let GeoJs render the other features?
  * I'm storing the color schemes (and I only have 3 of them) in multiple places.  does geojs expose the D3 or its own color schemes in some way, or should I just try to get them directly from D3?
  * The code to generate the clusters is fairly specific and brittle.  you can tie a color scheme to a dataset, but not to any features of the dataset.  and you can't change the size or coloration of a cluster from the fairly arbitrary choices I made.
  * The UI to change the geojson rendering is in the dataset panel, which is awkward when trying to change the rendering of a displayed dataset.  You need to remove it from the active session, change the property, then bring it back in.  This is relatively minor though.

jbeezley:
> * specifying the rendering of geojson makes sense assuming all you have is points, but what should the UI look like if your geojson is a mixture of points and other objects?
> * similarly, we either render as geojson using the geojs reader, or else we assume points and create the points from features. how can we render the other geojson objects other than the points if we have a geojson with a mix of features? look specifically in the MapPanel. Is there some way I can render the points as is but let GeoJs render the other features?

There is the geojson reader, but it is one of those simplified interfaces that probably won't be generic enough to do what you need.  It is pretty easy to parse a geojson object by looping through the [geometries](http://geojson.org/geojson-spec.html#geometry-objects) and passing the data to the appropriate feature type.  There are some complications in that you can have a geometry alone or as part of a feature.  A feature can also be part of a feature collection.  Maybe we can come up with a more general interface for allowing users to customize how it gets rendered.

> * I'm storing the color schemes (and I only have 3 of them) in multiple places. does geojs expose the D3 or its own color schemes in some way, or should I just try to get them directly from D3?

GeoJS doesn't expose any color schemes.  D3 is a prerequisite and should be available globally.

aashish24:
@mgrauer is this need s to be reviewed again?

mgrauer:
Thanks for checking @aashish24 .  I changed this to a WIP.  I'll pull some issues out of this and then close it.

mgrauer:
Closing this PR, as all functionality has been extracted or documented in issues.

PR #144, opened 2015-10-07
[WIP] Session fixes

mgrauer:
Some improvements to session endpoints, including a fix for an issue @jbeezley saw.

I've currently pushed a test that will exercise the failing behavior, I'll push a fix after the test fails.

### DIFF
--- a/plugin_tests/session_test.py
+++ b/plugin_tests/session_test.py
@@ -52,6 +52,9 @@ class SessionTestCase(base.TestCase):
         self._user = self.model('user').createUser(
             'minervauser', 'password', 'minerva', 'user',
             'minervauser@example.com')
+        self._nonAdminUser = self.model('user').createUser(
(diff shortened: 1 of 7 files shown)


mgrauer:
This PR has blown up (of course), as I'm trying to use Minerva as a non-admin user for the first time.  

:facepunch: 

I don't like the way I'm abusing some of the GET endpoints to create resources if they don't exist.  

What should happen is that upon a user creation event, we should create all necessary resources

    * user minerva folder
    * user dataset folder
    * user session folder
    * user should be added to the minerva group, which should have read access on the minerva collection, which is where analyses are stored
    * user should be added to the list of approved romanesco users (or perhaps we should handle our romanesco perms a different way)

aashish24:
@mgrauer are you going to update this one as well?

mgrauer:
will do

Issue #134, opened 2015-10-01
Write developer's documents for Minerva


aashish24:

mgrauer:
What should these be specifically @aashish24 ?  Say you had to pick 3 specific topics to start with...

aashish24:
* How to add new analysis to Minerva
* How to add new data sources and datasets in Minerva 

Would be good starting point.

kotfic:
These are both open topics related to developing use-cases for how to separate project specific requirements from Minerva as a product.

aashish24:
@kotfic sure. Just to be clear,I am not suggesting we go after this task but wanted to make sure that we have this in the backlog.

mgrauer:
Adding new sources is done.

Additional topics

  * how to create an analysis
  * how to create and run a new job
  * how to create endpoints

aashish24:
> Additional topics
> 
> how to create an analysis
> how to create and run a new job
> how to create endpoints

Also,
* how to add new view

PR #148, opened 2015-10-09
Elasticsearch source skeleton

mgrauer:
@danlamanna 

Can you QA this?  It should give you a good start.

Also nicks the corner of #134 .


aashish24:
:clap:

mgrauer:
I wanted to do one more thing, I'm done now.

cpatrick:
Ah [memories :smile_cat: ](http://www.cvedetails.com/cve/2014-3120).

danlamanna:
@mgrauer This looks great.

We agreed Elasticsearch as a source makes sense to have in core Minerva so others can use it. If it makes sense to you I'll implement a very rudimentary backend for ```minerva_elasticsearch_query``` and get this merged in.

This will require adding ```elasticsearch``` as a Minerva dependency, which is a pretty lightweight wrapper around ```urllib3```.

Following that, I will create analyses specific to the Elasticsearch indices people using Terra will want.

Sound good?

mgrauer:
This all sounds good to me.  I think it makes the most sense to add that additional endpoint to `elasticsearch_source.py`. 

I think we said that endpoint will do something like

  * create a Job (async local or Romanesco)
  * create a Dataset, tie it into the Job (there was some discussion about adding it to the meta/minerva object in the Job, we'll have to figure out the exact specifics of this later)
  * return the Job from the endpoint
  * in the async Job, bring back the query result, save it as a JSON file in the Dataset
  * create GeoJson from the JSON, as a collection of lat/long points, save the GeoJson file to the Dataset
  * set the minerva metadata similar to [this](https://github.com/Kitware/minerva/blob/bsve_to_geojson/server/jobs/bsve_search_worker.py#L108-L121)

And let me know if you have any better ideas or questions

danlamanna:
My concern with putting that into master is that in general, Elasticsearch indices won't always contain geospatial information. I want to avoid putting anything into core that is specific to a particular domain, Memex HT data in this instance.

So if we create the general endpoint which just processes an ES query and puts the resulting JSON into a dataset, then the user can do whatever they want with it.

Then in Terra, I will create a specific analysis that takes that dataset and alters it using the code you've mentioned.

mgrauer:
great, makes sense.

danlamanna:
@mgrauer This is ready for review, let me know if the general approach makes sense and feel free to be as nitpicky as you'd like :)

mgrauer:
I don't get the purpose of the elasticsearch dataset endpoint, it seems like it isn't used.  Can you remove it and the test?  Or else explain its use to me.

The way I see it, you have endpoints that do everything you need already, without this other endpoint

  * create an elasticsearch source
  * run a query job on an elasticsearch source
    * create a job
    * create a JSON dataset
    * the job runs the query on the source and uploads the result to the JSON dataset

danlamanna:
I'll remove it, I originally added it to follow convention with the WMS endpoints.

server/rest/elasticsearch_source.py
```diff
@@ -0,0 +1,133 @@
...
+
+        minerva_metadata = {
+            'dataset_id': dataset['_id'],
+            'source': 'elasticsearch',
+            'elasticsearch_params': elasticsearchParams,
+            'original_type': 'json'
+        }
+        metadata['minerva'] = minerva_metadata
+        self.model('item').setMetadata(dataset, metadata)
+
```
mgrauer: Can you replace

    if 'meta' in dataset:
    ...
    self.model('item').setMetadata(dataset, metadata)

with

    minerva_metadata = {
        'dataset_type': 'json',  
        'source_id': params['sourceId'],
        'source': 'elasticsearch',
        'elasticsearch_params': elasticsearchParams,
        'base_url': baseURL
    }
    updateMinervaMetadata(dataset, minerva_metadata)

?

I know this wasn't very clear.

server/rest/elasticsearch_source.py
```diff
@@ -0,0 +1,133 @@
...
+        self.model('job', 'jobs').scheduleJob(job)
+
+        return minerva_metadata
```
mgrauer: can you `return job` instead of `return minerva_metadata` ?

What I'm trying to push towards is a convention where if some async job is created by an endpoint, the endpoint returns the job.

server/rest/elasticsearch_source.py
```diff
@@ -0,0 +1,133 @@
...
+class ElasticsearchQuery(Resource):
+
+    def __init__(self):
+        self.resourceName = 'minerva_query_elasticsearch'
+        self.route('POST', (), self.queryElasticsearch)
+
+    @access.user
+    def queryElasticsearch(self, params):
+        """
+        Creates a dataset to store the results for an elastic search query,
```
mgrauer: Also adjust this docstring to be 

Creates a local job to run the elasticsearch_worker, the job will store the results of the elastic search query in a dataset.

mgrauer:
Will you want to convert this to a romanesco job?  If so, please add an issue for that.

danlamanna:
@mgrauer Updated. This look good to you?

mgrauer:
LGTM.

Issue #142, opened 2015-10-04
test encryption of credentials tied to wms source creation endpoint


mgrauer:
see #132 .

maybe we could test the encryption/decryption utilities at least, even if it is too tricky to test the encryption of credentials in the wms source creation endpoint.

PR #151, opened 2015-10-13
Unit test for WMS source authentication encryption/decryption

mbertrand:
For https://github.com/Kitware/minerva/issues/142

### DIFF
--- a/plugin_tests/wms_test.py
+++ b/plugin_tests/wms_test.py
@@ -90,8 +90,7 @@ class WmsTestCase(base.TestCase):
         Test the minerva WMS dataset API endpoints.
         """
 
-        # create the source
-
+        # Create the source.
         path = '/minerva_source_wms'
         name = 'testWMS'
         typeName = 'geonode:global_temp'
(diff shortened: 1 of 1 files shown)


mgrauer:
@mbertrand Your changes LGTM.  I've imported the decryptCredentials method and then tested it directly.

I've gotten hung up on those imports before also, from girder.plugins.minerva, inside the tests.  I think the imports need to come after all of the dependent plugins have been enabled, or else they will fail because the minerva plugin can't resolve its dependencies.  

Those plugins are enabled in the [setupModule method of the test](https://github.com/Kitware/minerva/blob/53c5fd23c35c3debb3f55be049d1862678e57bf9/plugin_tests/wms_test.py#L30-L38), so if you import in the actual test case it will work.  Sorry for the trouble there.  Can you QA my changes when you get a chance, assuming Travis passes?

mbertrand:
@mgrauer your changes LGTM,  thanks for the tips!

mgrauer:
I'll merge this as soon as tests pass.  It fell victim to the Travis issues last week.

:roller_coaster:

PR #154, opened 2015-10-14
RTD Fixes and auto-building

danlamanna:
This fixes some of the RTD errors I was having, and it now builds on RTD off of this branch. When merged I'll flip it to build off master.


kotfic:
LGTM

PR #155, opened 2015-10-15
Remove docs from readme

danlamanna:
@kotfic Do we want to remove the NEX specific documentation from the readme, or at least put it in the NEX analyses directory?

@aashish24 Is this a good time to add the license to the readme, are we going with Apache 2.0?


aashish24:
> @aashish24 Is this a good time to add the license to the readme, are we going with Apache 2.0?

Yes, Apache-2.0. Do you want to add it (we can just copy-paste from GeoJS) or should I? We also want a LICENSE file if it doesn't exist

aashish24:
@danlamanna thanks!

kotfic:
LGTM

PR #158, opened 2015-10-20
Add opacity layer controls and filter/search feature

essamjoubori:
- Ability to filter/search data sources
- Ability to control layer opacity


aashish24:
nice @essamjoubori

mgrauer:
@essamjoubori Is this ready for QA?  I've got a PR #159 that touches a lot of this same ground, so I'll have to merge them in together.

essamjoubori:
@mgrauer Yes it's ready for QA

mgrauer:
This looks really nice.

screen shot 2015-10-21 at 1 57 01 pm

I can filter the layers based on a search

screen shot 2015-10-21 at 1 57 52 pm

We may want a feature like "select all of the filtered layers", but I think not for now.  That kind of checkbox  multiselect of sets/subsets can get really hairy with many edge cases, and we don't have a clear need.  This will probably become more apparent with a large number of layers.


The opacity controls work for WMS, very cool!

screen shot 2015-10-21 at 2 01 18 pm

mgrauer:
QA

- [ ] rebase against master before repushing (or you can merge in master, but I would prefer to avoid so many merges of master)
- [ ] I want us to do a better job of namespacing css selectors, I know it is a mess now.  Be sure to update the jade, the javascript and the styles for each of these
  - [ ] .remove-dataset-from-layer => .m-remove-dataset-from-layer
  - [ ] .opacity-range => .m-opacity-range
  - [ ] #filter-layers => #m-filter-layers
  - [ ] .opacity-contanier => .m-opacity-container (note the spelling change for container also)
  - [ ] .opacity-range => .m-opacity-range
  - [ ] #filter-layers => #m-filter-layers
  - [ ] .wms-layers => .m-wms-layers
  - [ ] .item-list => .m-item-list (this might even be too generic, possibly better to have a more specific name)
- [ ] add a margin-top 10px to the search box, so it isn't quite as tight against the title text
- [ ] add in the poorly spelled `icon-ajust` to the opacity control, or another icon of your choosing
screen shot 2015-10-21 at 2 00 34 pm
- [ ] divs are the default tag in jade, so you don't need to declare them
- [ ] wms-layers style is declared in two places
- [ ] opacity breaks on anything but wms, probably easiest is to alter the jade
```
    if dataset.getDatasetType() === 'wms'
        .m-opacity-container
        input.m-opacity-range&attributes({'m-dataset-id': dataset.get('_id')})(type='range', min=0, max=1, step=0.01, value=dataset.get('opacity'))
```
- [ ] can you clean up the cursors and selection for the WMS layers? i.e., I'd like a "cursor pointer" on the checkboxes and the labels of the WMS layers, and I'd like to toggle the checkbox state if the checkbox or the label is clicked.  I did something similar [here](https://github.com/Kitware/minerva/commit/1a5119328ca12aa22def16e8e4715e2c85273a01).

danlamanna:
And nitpicks:
- I think there are a couple places ```_.bind``` is being used where it doesn't have to be.
- Are WMS datasets guaranteed to have an opacity attribute?

jbeezley:
> * Are WMS datasets guaranteed to have an opacity attribute?

The opacity is applied by geojs for an arbitrary WMS dataset.  The reason the opacity controls aren't being displayed for other layers is because geojs doesn't support it for other types.

aashish24:
@essamjoubori is this ready for review (other than merge conflicts?)

essamjoubori:
@aashish24 Yes it's ready for review

aashish24:
thanks @essamjoubori, this is a very cool feature! I updated the branch with current master before we review/merge it.

aashish24:
Dashboard is green. @mgrauer are you planning to review it?

mgrauer:
Yes, I'm planning to review it. Focused on BSVE SDK for now.

mgrauer:
- [ ] Seems like some old styles snuck back in, can you remove them?
    * `.wms-layers` in web_external/stylesheets/body/dataPanel.styl
    * `.opacity-contanier` in web_external/stylesheets/body/dataPanel.styl
    * `.wms-layers` and possibly the entire file of web_external/stylesheets/body/wmsLayers.styl

mgrauer:
The only other thing I found was that when a layer is already added as a dataset, when you display the WMS source, the line in the source tied to the layer is disabled and checked (correct behavior), the checkbox has a disabled type cursor on it, but the layer text has a `cursor: pointer`.

- [ ] Can you change the cursor over the layer text in this disabled case to be whatever the name of the deselected cursor style is (`not-allowed` possibly), the same as the one you are using (or getting by default) for the checkbox?

mgrauer:
@essamjoubori 

I (hopefully) fixed the wms tests, which were unrelated to your work here.  These tests rely on external resources, and this was likely to happen at some point.  My changes to the wms tests mock what I can mock and monkey patch the rest, and as I'm writing this I realized I also needed to add test data, so now I've added that as well.

I played around with the filtering work here, and I saw that if you had a layer already added as a dataset, if that same layer exists in a different wms_source, then you couldn't add the layer.  I pushed a change to fix that, since most of the work towards fixing it was identifying the problem anyway.

Can you QA these commits and let me know if they are OK?  I give everything an LGTM.

essamjoubori:
@mgrauer LGTM. Thanks!

mgrauer:
We have 2 tests failing in master due to the recent merge of #159.  This branch fixes one of the tests, the `wms_test`.  The other test that is failing, `web_client_minerva`, I believe is due to some recent changes in Girder that we may not be compatible with yet.  

Since master is broken, and merging this branch will leave master less broken, I'm going to merge this branch even with one failing test.

mgrauer:
To be clear, #159 didn't break any tests, merging any commits to master would have re-run tests on master, and re-running tests on master would have failed.

PR #166, opened 2015-10-23
WMS Feature Information Display

mbertrand:
Added a WMS Feature Info widget that makes 'GetFeatureInfo' calls to retrieve attributes for every WMS layer on the map. Activated by clicking mouse + Shift key anywhere on the map. The GetFeatureInfo currently requests responses in HTML table format. The jQuery UI library has been added for displaying the info in a dialog widget, let me know if you'd prefer not to use this in Minerva.


mbertrand:
Screenshot:

screen shot 2015-10-23 at 11 52 24 am

TODO's:
- Formatting improvements (customize attribute display, show only first few attributes by default, etc)
- Display legend vs actual values (not sure where this would go in UI)

danlamanna:
Additionally re jQuery UI: @aashish24 - if we do bring this in, this is probably what we should use for collapsible panels.

jbeezley:
I fixed the click event @https://github.com/OpenGeoscience/geojs/pull/456 if you want to try that for triggering the info panel.

mbertrand:
Thanks @jbeezley, I tried out the event-click geojs branch and the changes work great.  However, with both this branch and master branch I get a double-map when adding a new layer, they both have zoom controls and pan/zoom in sync with each other:

screen shot 2015-10-23 at 2 22 20 pm

jbeezley:
Does it help if you change the map container's css to `position: relative`  [here](https://github.com/Kitware/minerva/blob/master/web_external/stylesheets/body/mapPanel.styl#L15)?

mbertrand:
@jbeezley I changed the map container css to position:relative but that didn't help.  Changing position from relative to absolute [here](https://github.com/OpenGeoscience/geojs/blob/master/src/ui/uiLayer.js#L28) worked for me.

jbeezley:
I'm not sure I understand exactly what is going on.  I'll try to run it locally.  I think that the uiLayer should actually be `absolute`, but too often with css my workflow is to try things randomly until something works.

@danlamanna Do you remember if there was a specific reason that the [uiLayer](https://github.com/OpenGeoscience/geojs/blob/master/src/ui/uiLayer.js#L28) uses `position: relative`?

aashish24:
@mbertrand are you going to update this branch?

mbertrand:
@aashish Eventually I plan to change the feature info activation from
shift-click to just click,  but I think that perhaps should wait until
Jonathan's recent geojs enhancements addressing click issues makes it into
the version used by default in Minerva?

jbeezley:
@mbertrand , #167 should include the fixes.

mbertrand:
Changed shift-click to just click for WMS Info display

Gruntfile.js
```diff
@@ -153,6 +153,7 @@ module.exports = function (grunt) {
                 '/' + staticDir + '/geo.ext.min.js',
                 // '/' + rootStaticDir + '/libs.min.js', // libs included in jade template
                 '/' + staticDir + '/jquery.gridster.js',
+                '/' + staticDir + '/jquery-ui.min.js',
```
danlamanna: Let's install this using npm rather than putting the source in the codebase.

aashish24: > Let's install this using npm rather than putting the source in the codebase.

Sounds like a good idea +1 @mbertrand what you think?

mbertrand: Yes, working on it now.

mbertrand: @danlamanna, @aashish24 I noticed jquery-ui is already imported into geojs via bower, so I modified Gruntfile.js to copy them in from there.

aashish24: thanks @mbertrand, do you have to copy? Or can you not refer to it directly?

mbertrand: @aashish24 The convention right now is to copy required js/css files to a 'static' directory, including the geojs files.  I just added the jquery-ui files to the list.

jbeezley: Yes, the files have to be moved under the static web root to be served by cherrypy.

aashish24: I see, thanks @mbertrand @jbeezley.

I am planning to review this soon but it would be great if @mgrauer can review it as well.

mgrauer:
I will review this, but it'll be later in the week.

aashish24:
@mbertrand I checked out your branch, its looking very good :star: . Few observations though:

* Clicking on anywhere on the map view results in a pop up with no information. Either, we should not display a pop up when we didn't pick up any value (preferred) or it should say no data

* Would it be possible to have some transparency on the pop up? 

I didn't look at the actual code.

mbertrand:
@aashish I think both of these should be possible.

On Wed, Nov 4, 2015 at 11:47 PM, Aashish Chaudhary <notifications@github.com

aashish24:
@mbertrand do you want to push it on the same branch?

mbertrand:
sure

mbertrand:
Popup should no longer display when no results are returned, and the display widget is transparent.

aashish24:
thanks @mbertrand looks good to me @mgrauer ?

aashish24:
:+1: awesome feature

mgrauer:
We will need to keep an eye on #187 as this branch changes the Gruntfile and that one does also.

aashish24:
@mgrauer if the changes looks good to you why don't we merge this one and the #187 can be updated afterwards?

mgrauer:
Works for me.  I'm QAing this now--but in a bit of a QA recursion now.

mgrauer:
I merged master in a couple times from different PRs going in.

web_external/js/views/widgets/WMSFeatureInfoWidget.js
```diff
@@ -0,0 +1,110 @@
...
+
+    callInfo: function (layer_idx, coords) {
+        if (this.layers.length > 0) {
+            var url = this.getUrl(layer_idx, coords);
+            var layer_name = this.layers[layer_idx].layerName;
+            var panel = this;
+            $.ajax({
+                url: url,
+                jsonp: false,
+                async: false,
```
mgrauer: can we make this `async: true`?  Local testing shows me it works, but maybe there are other considerations?

web_external/js/views/widgets/WMSFeatureInfoWidget.js
```diff
@@ -0,0 +1,110 @@
...
+                jsonpCallback: 'getLayerFeatures',
+                error: function (a, b, c) {
+                    console.log(c);
```
mgrauer: console.error

web_external/js/views/widgets/WMSFeatureInfoWidget.js
```diff
@@ -0,0 +1,110 @@
...
+                }
+            }).done(function (data) {
+                //panel.content += data;
```
mgrauer: can we remove this commented line?

web_external/js/views/widgets/WMSFeatureInfoWidget.js
jbeezley: Also, why is this `jsonp: false` with a `jsonpCallback`?

web_external/js/views/widgets/WMSFeatureInfoWidget.js
```diff
@@ -0,0 +1,110 @@
...
+                if ('features' in data && data.features.length > 0) {
+                    var layer_div = document.createElement('div');
```
mgrauer: I would prefer it we could create all of the HTML in jade, perhaps some restructuring would be necessary to make the render happen upon the async return.  I'd rather you not change this now, because we are probably going to want to sit on this for a bit, think about it, and then decide on where to take it--we can rewrite it then.  If you have some specifics about why this couldn't be done in the jade, let me know.  I'll add an issue to revise this at a later date.

web_external/js/views/body/MapPanel.js
```diff
@@ -115,6 +134,10 @@ minerva.views.MapPanel = minerva.View.extend({
             this.legendWidget[datasetId].remove(datasetId);
             delete this.legendWidget[datasetId];
         }
+        var layerIndex = $.inArray(layer, this.map.featureInfoWidget.layers);
```
mgrauer: This breaks in this case:

  * refresh the page
  * add any non-wms dataset to the map
  * remove the non-wms dataset from the map

In this case, `this.map.featureInfoWidget` has not been created and the line above fails.

Can you create the featureInfoWidget in the MapPanel's init method?

It's looking like you had to go through some contortions to get this all working, and I doubt I understand them.  This might relate to how you wrote the featureInfoWidget.  We may need a longer discussion to revise.

web_external/templates/widgets/wmsFeatureInfoWidget.jade
```diff
@@ -0,0 +1,6 @@
+div(id='wms_info_dialog', title="WMS Layer Info")
```
mgrauer: I'd like us to start namespacing using this convention `m-wms-info-dialog`.

There are 3 uses of this in WMSFeatureInfoWidget.js also.

mgrauer:
@mbertrand I'm done my QA pass for now.  I bailed out a bit shallow because I think we need to talk in person a bit about what is going on here.

web_external/js/views/widgets/WMSFeatureInfoWidget.js
mbertrand: Made async: true and removed jsonpCallback

web_external/js/views/widgets/WMSFeatureInfoWidget.js
mbertrand: removed

web_external/js/views/widgets/WMSFeatureInfoWidget.js
mbertrand: removed

web_external/js/views/body/MapPanel.js
mbertrand: I moved this block of code inside the following conditional block that makes sure the layer being removed is actually a WMS layer, and also added a check to make sure the featureInfoWidget exists.

I can create the widget inside the MapPanel's init method, but my thinking was that it should be created only if a WMS layer is added to the map.  I'm happy to go with whichever option you prefer.

web_external/templates/widgets/wmsFeatureInfoWidget.jade
mbertrand: Renamed to m-wms-info-dialog

web_external/js/views/body/MapPanel.js
mgrauer: Thanks @mbertrand , let's keep it this way for now so we can get this in sooner.

mgrauer:
@mbertrand #187  went in to fix the build.

I'll merge that in here as part of my QA and push a commit.

mgrauer:
@mbertrand I've got this branch ready to go in, if it looks good to you and assuming tests pass.

LG from my end.

I had two layers loaded, clicked on Brazil, and this is what I got

screen shot 2015-11-10 at 5 21 50 pm
 
Pretty sweet!

mbertrand:
@mgrauer thanks for merging the latest from master in.  LGTM

Issue #145, opened 2015-10-08
have all jobs provide links to job outputs


mgrauer:
If a job creates a dataset or source or whatever, the job should provide a way of accessing the output, at least by the time the job succeeds.

Something like

```
output_type: [dataset|source]
source_id: id
```

This way, when we listen for the job notifications in the web client, we can load the individual output and add it to the correct collection, rather than refetching the collection, which causes us to lose state.

mgrauer:
I'm going to try namespacing this within the kwargs of the job under minerva, e.g.

```
kwargs = {
    ...
    'minerva': {
        'output_type': 'dataset',
        'dataset_id': dataset['_id']
    }
}
```

mgrauer:
Actually, we should do this under the meta object on the job, in a minerva namespace.

mgrauer:
If a job fails during processing, it should delete whatever resources it is attached to, i.e. a source or dataset that it has created and modified.

mgrauer:
Currently there is a bug where

  * if you have datasets that are currently displayed in the session
  * then you create a new dataset via a job
  * when the job finishes, the dataset collection will be fetched, which erases the `displayed` state in the shown datasets
  * and then if any additional dataset is requested to be shown, the previously shown datasets will continue to be shown in the map panel, but they will disappear from the session panel display

A fix for this is to have the job status listeners on the data and session panel listen for a job success event of the correct type, then--because we will be storing the job output pointer as detailed above--we should load that individual resource model and add it to the appropriate collection type, then rerender the corresponding list panel.  That way the collection won't be fetched, and our new resource will show up in the appropriate list (and we can even do some fancy UI highlighting on the new resource row if desired).

aashish24:
So how its done right now?

mgrauer:
We listen for job status updates on the [Dataset Panel](https://github.com/Kitware/minerva/blob/master/web_external/js/views/body/DataPanel.js#L105-L110) and coming in with #159, the [Source Panel](https://github.com/Kitware/minerva/blob/S3_source_refactor/web_external/js/views/body/SourcePanel.js#L126-L132).  If the job status is `success`, we do a fetch on the dataset/source collection.

aashish24:
> We listen for job status updates on the Dataset Panel and coming in with #159, the Source Panel. If the job status is success, we do a fetch on the dataset/source collection.

[...]

Issue #160, opened 2015-10-20
Add ability for SourceModel subclasses to register themselves


mgrauer:
#159 provided some prefactoring to make it easier to add a new SourceModel subclass.

When adding a new SourceModel subclass a developer still needs to

  * add the source_type to the SourceCollection.model function switch
  * add the icon action event handler to SourcePanel and specify a widget
  * add itself to the dictionary of sourcetypes held in SourcePanel
  * add its creation widget to the dictionary in AddSourceWidget

Ideally some registration method should be provided that can set all of these properties at one time.

Maybe a method on the minerva application object?  And when a developer defines a new SourceModel subclass, in that same source file there could be an immediate that registers the subclass with the minerva application object.  This provides some ability to define client side code outside of minerva and have it imported into minerva simply by enabling another girder plugin (assuming minerva's core template loads only client side code from enabled plugins, which is a good change anyway).

I'd love to hear any other good ideas or potential problems with this approach.

mgrauer:
A required part of this will be updating the "add a source" documentation.

danlamanna:
I think this is the sort of approach we've been thinking about for all of the client-side code in Minerva (read: plugin system), not just source models.

I think using the Minerva application object as a sort of registry makes sense. In fact, other girder plugins could have their own gruntfile namespacing jade/stylus files as ```my_own_plugin.templates``` so all the client-side code could be self-contained.

aashish24:
> I think this is the sort of approach we've been thinking about for all of the client-side code in Minerva (read: plugin system), not just source models.

Agree. @mgrauer your approach sounds good to me. Having registry within the minerva application object is good as well. I don't have much to add.

mgrauer:
I agree this is desirable for all the client-side code, but sources are closest to having the api defined cleanly enough.  What's nice is by making it easier to add new sources within Minerva this way, we also enable the ability from plugins.

Analyses would probably be next.  

Datasets would be harder, as you'd have to also add some ability to render a new type that Minerva wasn't aware of.  Changes to support #157 can enable this.

aashish24:
> Analyses would probably be next.

I was thinking the same that Analysis can be dealt in very similar way. 

> Datasets would be harder, as you'd have to also add some ability to render a new type that Minerva wasn't aware of. Changes to support #157 can enable this.

[...]

PR #174, opened 2015-11-02
Bsve analysis update

mgrauer:
I've based this on top of the S3 source refactor, and can rebase once that gets into master.

This pulls out and cleans up the bsve search analysis work from #143--that is, calling the bsve search API, getting the results as json, converting them to lat/long geojson points, and setting the dataset output to be geojson.

Also follows the path laid out in #145, in that it creates links to the dataset as output of the job, stored in the job's metadata's minerva namespace.

It further updates the tests to be in sync with the code.

### DIFF
--- a/server/rest/s3_source.py
+++ b/server/rest/s3_source.py
@@ -38,7 +38,7 @@ class S3Source(Source):
         user = self.getCurrentUser()
         name = params['name']
         bucket = params['bucket'].strip()
-        prefix = params['prefix'].strip()
+        prefix = params.get('prefix', '').strip()
         access_key_id = params.get('accessKeyId', '').strip()
         secret = params.get('secret', '').strip()
         service = params.get('service', '').strip()
--- a/web_external/js/views/widgets/AddS3SourceWidget.js
+++ b/web_external/js/views/widgets/AddS3SourceWidget.js
@@ -22,11 +22,19 @@ minerva.views.AddS3SourceWidget = minerva.View.extend({
                 params.prefix = '/' + params.prefix;
             }
 
+            if (name.trim() === '' || bucket.trim() === '') {
+                this.$('.g-validation-failed-message').text('Source name and S3 bucket name are required');
+                return;
+            }
+
             var s3Source = new minerva.models.S3SourceModel();
             s3Source.on('m:sourceReceived', function () {
                 this.$el.modal('hide');
+                this.$('.g-validation-failed-message').text('');
                 // TODO: might need to be added to a new panel/data sources ?
                 girder.events.trigger('m:job.created');
+            }, this).on('m:error', function (msg) {
+                this.$('.g-validation-failed-message').text(msg.responseText);
             }, this).createSource(params);
         }
     },
--- a/web_external/js/views/widgets/AddS3SourceWidget.js
+++ b/web_external/js/views/widgets/AddS3SourceWidget.js
@@ -22,7 +22,7 @@ minerva.views.AddS3SourceWidget = minerva.View.extend({
                 params.prefix = '/' + params.prefix;
             }
 
-            if (name.trim() === '' || bucket.trim() === '') {
+            if (params.name.trim() === '' || params.bucket.trim() === '') {
                 this.$('.g-validation-failed-message').text('Source name and S3 bucket name are required');
                 return;
             }
--- a/web_external/js/views/widgets/ReadOnlyHierarchyWidget.js
+++ b/web_external/js/views/widgets/ReadOnlyHierarchyWidget.js
@@ -97,7 +97,7 @@ minerva.views.ReadOnlyHierarchyWidget = girder.views.HierarchyWidget.extend({
      * the checked menu state.
      */
     updateChecked: function () {
-        // do nothing
+        // No Op, the superclass requires an implementation.
     },
 
     getCheckedResources: function () {
--- a/plugin_tests/analysis_test.py
+++ b/plugin_tests/analysis_test.py
@@ -160,13 +160,18 @@ class AnalysisTestCase(base.TestCase):
                 },
                 user=self._user
             )
+            self.assertStatusOk(response)
+            job = response.json
 
             # wait for the async job to complete
             searchResultsFinished = False
             count = 0
+            output = job['meta']['minerva']['outputs'][0]
+            self.assertEquals(output['type'], 'dataset', 'Incorrect output type %s' % output['type'])
+
             while not searchResultsFinished and count < 5:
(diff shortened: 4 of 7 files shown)


web_external/js/MinervaModel.js
```diff
@@ -9,6 +9,14 @@ minerva.models.MinervaModel = girder.models.ItemModel.extend({
         }
     },
 
+    mM: function (minervaMetadata) {
+        if (minervaMetadata) {
```
danlamanna: Can we rename this to ```minervaMetadata``` or something that follows suit with the rest of the class?

server/rest/s3_source.py
```diff
@@ -0,0 +1,120 @@
...
+        self.resourceName = 'minerva_source_s3'
+        self.route('POST', (), self.createS3Source)
+
+    @access.user
+    def createS3Source(self, params):
+        assetstore = self.model('assetstore')
+        user = self.getCurrentUser()
+        name = params['name']
+        bucket = params['bucket'].strip()
+        prefix = params['prefix'].strip()
```
danlamanna: This will throw a KeyError if someone doesn't pass in a prefix.

mgrauer: Good point, though this is merged into master already.  I can fix it here.

web_external/js/MinervaModel.js
mgrauer: My motivation was that I got sick of typing out getMinervaMetadata/setMinervaMetadata all the time.  This is something that will be called often throughout the application.

Is there some other shorthand you can live with?  

`meta`?
`metadata`?
`mMetadata`?
`minerva`?

I'm willing to go with `minervaMetadata` if you can't stomach anything else.  :mask:

web_external/js/views/widgets/AddS3SourceWidget.js
```diff
@@ -0,0 +1,46 @@
...
+                secret: $('#m-s3-secret').val(),
+                service: $('#m-s3-service').val(),
+                readOnly: $('#m-s3-readonly').val(),
+                folderId: this.collection.folderId
+            };
+
+            if (params.prefix.indexOf('/') !== 0) {
+                params.prefix = '/' + params.prefix;
+            }
+
```
danlamanna: Can we do some primitive client-side validation here similar to [this](https://github.com/Kitware/minerva/blob/bsve_analysis_update/web_external/js/views/widgets/ElasticsearchWidget.js#L24)?

web_external/js/views/widgets/AddSourceWidget.js
```diff
@@ -6,39 +6,63 @@ minerva.views.AddSourceWidget = minerva.View.extend({
         'submit #m-add-source-form': function (e) {
             e.preventDefault();
 
+            // TODO better to have single buttons instead of radio?
+            // That way user only has to click once instead of radio + Add
```
danlamanna: I agree, in fact I don't even think this necessarily has to be a widget. Rather, the + icon could just trigger a dropdown of clickable source types.

Simplifying/changing that can be it's own issue though.

web_external/js/views/widgets/AddSourceWidget.js
```diff
@@ -6,39 +6,63 @@ minerva.views.AddSourceWidget = minerva.View.extend({
...
+                icon: 'icon-search',
+                widget: minerva.views.AddElasticsearchSourceWidget
+            },
+            s3: {
+                label: 'S3 bucket',
+                icon: 'icon-cloud',
+                widget: minerva.views.AddS3SourceWidget
+            }
+        };
     },
```
danlamanna: Since these are going to be in the SourceCollection and they're hardcoded, does it make sense to put them there now? I understand this is related to their UI, but it seems pretty counter-intuitive that the types of sources are registered in a widget view.

This way we could add the source model attribute to this definition and remove the if/else statements in SourceCollection.

This might not make sense since we're planning to refactor a lot of this soon anyway, in which case - disregard.

web_external/js/views/widgets/MeanContourWidget.js
```diff
@@ -64,18 +78,61 @@ minerva.views.MeanContourWidget = minerva.View.extend({
...
+                folderCreate: false,
+                folderEdit: false,
+                itemCreate: false
+            }).render();
+        }, this).on('g:error', function () {
+            // TODO
+            console.error('fetch of folder failed');
+        }, this).fetch();
+    },
+
```
danlamanna: I can see this function being useful in other widgets - maybe we should start thinking about creating a generic AnalysisWidget? #separateIssue

web_external/js/views/widgets/ReadOnlyHierarchyWidget.js
```diff
@@ -102,7 +97,7 @@ minerva.views.ReadOnlyHierarchyWidget = girder.views.HierarchyWidget.extend({
      * the checked menu state.
      */
     updateChecked: function () {
-        // do nothing - DatasetHierarchyWidget will handle updating the dataset
+        // do nothing
```
danlamanna: Can we just remove this then?

mgrauer:
@danlamanna Romanesco is a dependency.

https://github.com/Kitware/minerva/blob/2771a7a3620bb1d955e8460628c439ba92cbe098/plugin.json#L8

https://github.com/Kitware/minerva/blob/2771a7a3620bb1d955e8460628c439ba92cbe098/requirements.txt#L8

Also note that anything Romanesco related must be on line 8.

:8ball:

danlamanna:
touché.

Other than the remarks I had, LGTM.

web_external/js/MinervaModel.js
danlamanna: I could live with ```metadata```.

web_external/js/views/widgets/AddSourceWidget.js
mgrauer: I added a reference to this comment in the refactor issue, #160.

web_external/js/views/widgets/AddSourceWidget.js
mgrauer: Added #194 to capture this.

web_external/js/MinervaModel.js
mgrauer: Captured in #195.

web_external/js/views/widgets/MeanContourWidget.js
mgrauer: Captured in #196.

mgrauer:
@danlamanna PTAL.

danlamanna:
This LGTM.

PR #175, opened 2015-11-03
[WIP] Map view improvements

aashish24:
* Updated to use parallel and discreet zoom
* Using nicer looking tiles

### DIFF
--- a/web_external/js/views/body/MapPanel.js
+++ b/web_external/js/views/body/MapPanel.js
@@ -150,7 +150,9 @@ minerva.views.MapPanel = minerva.View.extend({
             this.map = geo.map({
                 node: '.mapPanelMap',
                 center: this.session.sessionJsonContents.center,
-                zoom: this.session.sessionJsonContents.zoom
+                zoom: this.session.sessionJsonContents.zoom,
+                parallelProjection: true,
+                discreteZoom: true
             });
             this.map.createLayer(this.session.sessionJsonContents.basemap);
             this.uiLayer = this.map.createLayer('ui');
--- a/web_external/js/views/body/MapPanel.js
+++ b/web_external/js/views/body/MapPanel.js
@@ -154,7 +154,8 @@ minerva.views.MapPanel = minerva.View.extend({
                 parallelProjection: true,
                 discreteZoom: true
             });
-            this.map.createLayer(this.session.sessionJsonContents.basemap);
+            this.map.createLayer(this.session.sessionJsonContents.basemap,
+              {baseUrl: 'http://otile1.mqcdn.com/tiles/1.0.0/map/'});
             this.uiLayer = this.map.createLayer('ui');
             this.uiLayer.createWidget('slider');
             this.mapCreated = true;


jbeezley:
I wouldn't use discrete zoom until we fix the [issue](https://github.com/OpenGeoscience/geojs/issues/424) with touchpads.

aashish24:
@jbeezley supporting touch pads would be nice but currently we are not running our client on touchpads. I think using touch pads properly would require some UI work as well. Please let me know if I am missing something.

jbeezley:
Not touchscreen, touchpads like most laptops have.

aashish24:
> Not touchscreen, touchpads like most laptops have.

thanks @jbeezley. I remember we have a issue for in in GeoJS. Do we have any estimate on how long its going to take to fix that? I need to capture some screenshots for Minerva, so I can do it locally. @mgrauer should we wait before we fix the touchpad issue.

jbeezley:
I could do it in a few hours probably; somebody just needs to make it a priority.  What makes it nontrivial is that we need a way to debounce mouse wheel events while accumulating the total movement.  It would be helpful for performance as well.

Merging as is really isn't an option.  Discrete zoom makes the map basically unusable on a laptop.

aashish24:
I see, so we need to debounce mouse wheel events so that it does not fire so often? 

> I could do it in a few hours probably; somebody just needs to make it a priority.
It is now :+1: 

> Merging as is really isn't an option. Discrete zoom makes the map basically unusable on a laptop.
Sure, this branch can wait until that happens

jbeezley:
Yes, touchpads and other devices fire fast wheel events with a small delta usually in units of pixels.  Mouse wheels that "click" give discrete events typically in units of lines.

aashish24:
> Yes, touchpads and other devices fire fast wheel events with a small delta usually in units of pixels. Mouse wheels that "click" give discrete events typically in units of lines.

this is informative. thanks @jbeezley

aashish24:
@mgrauer once we update the geojs, I will rebase this branch.

aashish24:
@jbeezley is touchpad working now? I haven't tried it on Mac yet

jbeezley:
You can try for yourself here: http://opengeoscience.github.io/geojs/examples/tiles/?discrete=true.

My personal opinion is that the it feels awkward and something like Google Map's rubberband behavior is necessary for a positive user experience.  Perhaps others can try it out offer feedback as well.

aashish24:
> My personal opinion is that the it feels awkward

Yeah, I tried it yesterday and I agree. The experience was not great. Is it because of performance or something else? 


> Google Map's rubberband behavior is necessary for a positive user experience

I am okay with having discreet zoom without this one for now but I agree a rubberband would be really awesome. Let's make sure that we have this in the geojs issue tracker. 

what I will do in the mean time is I will enable the parallel projection only so that alteast we will get better image when we are very close to the integer zoom level.

jbeezley:
Parallel projection is now the [default](https://github.com/OpenGeoscience/geojs/blob/master/src/core/camera.js#L791).  That syntax for setting the camera projection also isn't supported anymore.  Now this is done via: `map.camera().projection = 'parallel'` or `map.camera().projection = 'perspective'`.

aashish24:
> Parallel projection is now the default.

is that in release 0.7? I guess I didn't see difference when we were doing the demo but it could be because of the screen resolution of the projector. 

> That syntax for setting the camera projection also isn't supported anymore. Now this is done via: map.camera().projection = 'parallel' or map.camera().projection = 'perspective'.

Yeah I was going to update my current branch with the newer syntax. Also, will try to change the default WMTS.

jbeezley:
The default projection changed with 0.6.

aashish24:
> The default projection changed with 0.6.

thanks @jbeezley I am going to close this PR and will push another one, for now, I will just change the wmts entry then.

PR #187, opened 2015-11-07
Refactor of minerva build scripts

jbeezley:
To go with https://github.com/girder/girder/pull/1065, I am working on a similar update here to tame our build process a bit.  I'm not finished with it, but I wanted to give heads up to anyone else that might be thinking of making changes to `Gruntfile.js`.

One thing this PR will do is make it so that you don't need to run any commands from inside Minerva's directory.  You will just need to do `grunt init && grunt` within Girder's directory as usual.  I will probably need to make a few more changes to Girder to get it working. 

Also, I haven't touch the build task `test-env-html:minerva`, which is pretty unruly.  Ultimately, I think it would be best to build it with webpack, but that is a more ambitious task that I will save until later.


aashish24:
:clap:

jbeezley:
@mgrauer The client test was failing because the test html file was compiled before the static path variables were set.  It is an easy fix because you now specify task dependencies.  This *should* fix this build as well as the failure in #189.

mgrauer:
@jbeezley You rock!  Thanks, this worked for me, with a totally fresh clone of Minerva no less.

Is this still WIP?  LGTM if not.

jbeezley:
No, I'm done with it.  Merging now.

PR #197, opened 2015-11-11
Metadata unification

### LINKED ISSUE #195: Refactor client side minerva metadata
There are a few methods used by minerva models that should be unified into a `metadata` method.

  * mM
  * setMinervaMetadata
  * getMinervaMetadata

See [this comment](https://github.com/Kitware/minerva/pull/174#discussion_r44328436) and those above it.

mgrauer:
Fixes #195.

Based on top of commits from #174.

ping @danlamanna

### DIFF
--- a/server/jobs/bsve_search_worker.py
+++ b/server/jobs/bsve_search_worker.py
@@ -12,6 +12,7 @@ from girder.plugins.jobs.constants import JobStatus
 from girder.plugins.minerva.utility.bsve.bsve_utility import BsveUtility
 from girder.plugins.minerva.utility.dataset_utility import \
     jsonArrayHead, GeoJsonMapper, jsonObjectReader
+from girder.plugins.minerva.utility.minerva_utility import mM
 
 import girder_client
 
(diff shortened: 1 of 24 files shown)


server/jobs/bsve_search_worker.py
```diff
@@ -88,12 +86,42 @@ def run(job):
...
+            minerva_metadata['geojson_file'] = {
+                '_id': existing['_id'],
+                'name': geojsonFilename
+            }
+        else:
+            raise (Exception('Cannot find file %s in dataset %s' %
+                   (geojsonFilename, datasetId)))
+
+        mM(dataset, minerva_metadata)
         job_model.updateJob(job, status=JobStatus.SUCCESS)
```
danlamanna: I know calling it just ```metadata``` here would be ambiguous, but can we find some way to rename ```mM``` server-side too?

mgrauer: Opened issue #198.

mgrauer:
Was that your only comment @danlamanna , or did your QA get interrupted?  Just wondering where this PR stands.

danlamanna:
Yes, otherwise LGTM.

PR #200, opened 2015-11-12
Simplify travis npm to be consistent with install docs

mgrauer:


mgrauer:
@jbeezley PTAL.  This will be quick.

jbeezley:
LGTM.

There is a reason why one might want to update npm though.  I think at the time this was added in girder, the version of npm on travis was broken for some reason.  There were also times when the latest version of npm was broken (specifically right after 3.0 came out).  For now, I think `2.x` and `3.x` are both relatively safe for the dependency set we use.

Because npm needs to be globally installed, I don't know that it is a great idea to try to enforce a specific version on the users, and I definitely would advise against always upgrading to the latest.  As far as documentation is concerned, my vote would be to advise about versions that are known to be broken if/when we come across them.  For now, avoiding the issue in the docs is probably okay.

mgrauer:
Thanks for the detailed info.

PR #201, opened 2015-11-13
Load resources from all enabled plugins

mgrauer:
This enables other plugins to extend Minerva on the client side but still have Minerva be served as the main web application.

### DIFF
--- a/server/loader.py
+++ b/server/loader.py
@@ -18,12 +18,12 @@
 ###############################################################################
 
 import mako
-import json
 import os
 import requests
 import cherrypy
 from base64 import b64encode
-from girder import constants, events
+from girder import events
+from girder.constants import SettingKey, STATIC_ROOT_DIR
 from girder.utility.model_importer import ModelImporter
 
 from girder.plugins.minerva.rest import \
@@ -33,7 +33,7 @@ from girder.plugins.minerva.rest import \
 from girder.plugins.minerva.utility.minerva_utility import decryptCredentials
 
 
-class CustomAppRoot(object):
+class CustomAppRoot(ModelImporter):
     """
     The webroot endpoint simply serves the main index HTML file of minerva.
     """
@@ -123,17 +123,13 @@ src="http://cdn.jsdelivr.net/bootstrap.daterangepicker/1/daterangepicker.js">
         if self.indexHtml is None:
             self.vars['pluginCss'] = []
             self.vars['pluginJs'] = []
-            builtDir = os.path.join(constants.STATIC_ROOT_DIR, 'clients',
+            builtDir = os.path.join(STATIC_ROOT_DIR, 'clients',
                                     'web', 'static', 'built', 'plugins')
-            # it would be nice to get the activated plugins from girder's
-            # settings # but we need to reproduce this functionality in the
-            # Gruntfile, so pull these from the plugin.json
-            minervaPluginDir = os.path.dirname(os.path.realpath(__file__))
-            pluginJson = os.path.join(minervaPluginDir, '..', 'plugin.json')
-            with open(pluginJson, 'r') as pluginJsonData:
-                pluginConfig = json.load(pluginJsonData)
-                self.vars['plugins'] = pluginConfig['dependencies']
-            for plugin in self.vars['plugins']:
+            # Load resources from all enabled plugins, which allows Minerva
+            # to be extended by other Girder plugins.
+            plugins = self.model('setting').get(SettingKey.PLUGINS_ENABLED, ())
+            plugins.remove('minerva')
+            for plugin in plugins:
                 if os.path.exists(os.path.join(builtDir, plugin,
                                                'plugin.min.css')):
                     self.vars['pluginCss'].append(plugin)
--- a/server/loader.py
+++ b/server/loader.py
@@ -17,13 +17,13 @@
 #  limitations under the License.
 ###############################################################################
 
-import mako
 import os
 import requests
 import cherrypy
 from base64 import b64encode
 from girder import events
-from girder.constants import SettingKey, STATIC_ROOT_DIR
+from girder.constants import SettingKey
+from girder.utility.webroot import Webroot
 from girder.utility.model_importer import ModelImporter
 
 from girder.plugins.minerva.rest import \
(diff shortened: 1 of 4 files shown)


aashish24:
@mgrauer can you explain this bit more?

jbeezley:
I believe it just loads static content from all enabled plugins rather from a predefined list of plugins.

I'm wondering if it would be better just to use the recent change to girder core that moves the mako template to a separate file.  There is a keyword arg that lets you override the template used.

https://github.com/girder/girder/blob/master/girder/utility/webroot.py#L79

aashish24:
> I believe it just loads static content from all enabled plugins rather from a predefined list of plugins.

My wish list would be to have Minerva load a predefined list of plugins and that would enable the static content?

mgrauer:
The intention was to allow other plugins to extend Minerva, and not have to have Minerva know which ones are extending it, but still serve their client side resources.  Minerva would still work and serve the client side resources of all of its dependencies (defined in the plugin.json file), but if some other plugin wants to use Minerva and pass in for example a configurable UI, that plugin depends on Minerva in its plugin.json, but Minerva is unaware of the other plugin.  All of the client resources for that are other plugin are still served, even though the mako template is Minerva's.

I'll look into that mako template change you mentioned @jbeezley .

@aashish24 does explanation fulfill your wish?

aashish24:
> @aashish24 does explanation fulfill your wish?

this sounds great @mgrauer thanks!  do we have a use-case right now?

mgrauer:
Yes, for one of our other projects that involves a map plus a time series analysis that @danlamanna  is working on.

In the coming weeks @danlamanna will work towards the configurable layout milestone, so that his project's application can use this functionality to set its own layout and panels.

aashish24:
Cool! thanks @mgrauer. We actually need to bring some of the pieces from that app into Minerva proper as we may need to wrap up our work there. I can provide more detail tomorrow.

mgrauer:
@jbeezley I split out the mako template as you suggested.  Tests were failing, but I was able to get it down to only the web client test failing.

As a sanity check, I pulled dc63cd2 into a separate PR, #203.  That single commit, which doesn't change anything related to mako, had passed earlier on this PR this past Friday, although now it shows up as failing on this PR.  Every test on that PR 203 passes except for the web client test, so my guess is that the web client test is failing because of some change in Girder from Friday to today.

I can work with you and @zachmullen to run this down tomorrow.

mgrauer:
There's been a lot of great stuff coming into Girder lately, but we've had some broken builds.  I definitely want all the goodness related to better Grunt building and easier plugin installs, and we should probably adapt our Travis to using that.

I'm wondering if it would be better if we tie ourselves to the next minor release of Girder for the purpose of Travis.  The advantage is that we won't have to fix any more broken builds incrementally due to upstream changes, the disadvantage is that we may have a large chunk of work to do at the subsequent release.  Any strong opinions?

ping @Kitware/ess

jbeezley:
:+1:

jbeezley:
Also RE the client test, I tried to improve the grunt spec for building the `testEnv.html` but I found just about anything I did broke it.  The test file in this repo is significantly more complicated.  Ideally, the build task for this file should use `grunt-contrib-jade` with as much as possible of the template data provided statically (possibly using underscore-style templates) in the grunt config.

kotfic:
+1 for pegging to girder minor version

aashish24:
> I'm wondering if it would be better if we tie ourselves to the next minor release of Girder for the purpose of Travis. The advantage is that we won't have to fix any more broken builds incrementally due to upstream changes, the disadvantage is that we may have a large chunk of work to do at the subsequent release. Any strong opinions?

+1 :+1:

mgrauer:
Added an issue for pegging to minor version, #204.

mgrauer:
@jbeezley PTAL.

server/loader.py
```diff
@@ -168,8 +55,27 @@ def validate_settings(event):
...
+    plugins = list(plugins)
+    if 'minerva' in plugins:
+        plugins.remove('minerva')
+    vars = {
+        'plugins': plugins,
+        'apiRoot': '/api/v1',
+        'staticRoot': '/static',
+        'title': 'Minerva'
+    }
+    minerva_webroot.updateHtmlVars(vars)
```
jbeezley: A couple of issues here.

1. This is called [here](https://github.com/girder/girder/blob/master/girder/utility/plugin_utilities.py#L79) after girder core has determined what plugins are enabled, so none of the changes made here make a difference at least for the purpose of plugin loading.
2. Hard coding the api and static root like this will break a lot of deployment scenarios.

I would suggest simplifying this part to do something like the following:
```python
minerva_mako = os.path.join(os.path.dirname(__file__), "minerva.mako")
minerva_webroot = Webroot(minerva_mako)
minerva_webroot.updateHtmlVars(info['serverRoot'].vars)
minerva_webroot.updateHtmlVars({'title': 'Minerva'})
```

mgrauer: 1.  I think you are talking about "plugin loading" meaning on the server side, whereas I'm only intending this to have an effect on the client side resources of the loaded plugins, i.e. to have them be served in the minerva mako template.

If this is the case, it is probably a matter of me clarifying the comments here to better describe that this will only affect client side resources.  Do you agree or am I missing something else?

2. Will do.  Can you give an example of such a scenario so I can understand the issue better?

jbeezley: > I think you are talking about "plugin loading" meaning on the server side, whereas I'm only intending this to have an effect on the client side resources of the loaded plugins, i.e. to have them be served in the minerva mako template.

Correct, but more importantly I don't think there is any way that this plugin list can be different from the list that girder is [already initializing](https://github.com/girder/girder/blob/master/girder/utility/server.py#L103).  My code above uses the list girder is using for it's own web root (as well as the static and api roots).

> Will do. Can you give an example of such a scenario so I can understand the issue better?

http://girder.readthedocs.org/en/latest/deploy.html#reverse-proxy

zachmullen: > Will do. Can you give an example of such a scenario so I can understand the issue better?

[This section](http://girder.readthedocs.org/en/latest/developer-cookbook.html#serving-a-custom-app-from-the-server-root) of the docs discusses moving the girder webroot and the API root within cherrypy, which is another reason the API might not be at `/api/v1`. The static root may point, for instance, to a CDN, or get moved by a plugin for whatever reason.

Reading these vars from the core template is probably a good solution for now.

mgrauer:
@jbeezley I think this resolves the issues.

jbeezley:
LGTM

PR #206, opened 2015-11-18
Mongo Source/Dataset support

mbertrand:
Load MongoDB sources (databases) and datasets (collections).  Detects mappable data by looking for a spatial index ('2dsphere') or a 'geometry' attribute in the collection, and if found, creates a geojson object from the collection and saves it in the dataset's metadata.  If not found, allows the user to specify lat/long fields via the old table-mapper functionality formerly found in Minerva.  Brings back some other old minerva code as well to display the data on the map.


mbertrand:
screen shot 2015-11-18 at 12 20 32 pm
screen shot 2015-11-18 at 12 20 45 pm
screen shot 2015-11-18 at 12 20 58 pm
screen shot 2015-11-18 at 12 22 04 pm
screen shot 2015-11-18 at 12 22 18 pm

jbeezley:
ohshiny

aashish24:
Very nice @mbertrand two awesome features in a single day :smile:

aashish24:
:+1: for @jbeezley img, that's one of the best I have seen.

mgrauer:
I rebased this on master and may push some more commits as part of QA.

mgrauer:
The endpoint for mongo dataset creation was creating an additional item.  I've also removed the `dataset_id` property from the metadata.  My reasons for moving away from that because it is tracked in the item itself, it would be hard to keep all those consistent if items were copied, and I no longer need that in the metadata because I'm having dataset creation endpoints return the full item instead of just the minerva_metadata.

TODO: document these practices in dev guide.

mgrauer:
@mbertrand Can you add some datasets and tests that exercise the 2dsphere and geometry properties/paths through the code?  An extract of a handful of rows would be nice, and it would allow me to play around with that dataset.

mbertrand:
@mgrauer sure, will do

mgrauer:
In the interest of not adding more stuff that works only part of the time, could you remove the Date filter from the key mapping widget?  If we followed this policy even further, it would suggest that we should change the keymapper to something a bit more specific, since it only allows you to map to GeoJson points and it would show up even if not appropriate--but I can live with this for now since it at least works in the general case that your data has lat/long.

mgrauer:
Your source is for a Mongo database, and then you create datasets for individual collections.  If the dataset has GeoJson in its metadata, which would come from a conversion of every row in the collection to GeoJson, it can be visualized on the map.  You do allow in the GeoJson conversion for a subquery of the Mongo collection, but your dataset creation endpoint does not use a subquery.  Also, your GeoJson is not saved in the metadata of the dataset.

All that is just describing the behavior in this PR so we can have a wider discussion, though I don't think you should change this PR.

We've had a revision in Minerva's history already in splitting out Datasets into Sources and Datasets.  I think we need to go a bit further, but also need to come to an agreement on what exactly a Source, Dataset, and (new term, perhaps not the right word) Visualization are.  This should be broken out into one or several issues.

Here is my rough thinking on the subject, with what changes I would see related to Mongo sources if it were implemented.

**Source**

  * a source of data
  * an Item in Girder, living under a User's Minerva/Source folder
  * Minerva metadata on the Item informs where/what/how to get data

In this case

  * *what* JSON, or possibly "rows/JSON", since the semantics of the data isn't related to the format
  * *where* connection info for the Mongo DB/collection
  * *how* Mongo

I'm a bit torn as to whether a Mongo DB or Collection should be a source.

**Dataset** 

  * a static collection of data, the static is important to Gumbo because of reproducibility
  * an Item in Girder, living under a User's Minerva/Dataset folder
  * the bytes of the data would live in one or more Files in the Item
  * Minerva metadata on the Item informs the type of data
  * Provenance would be tracked on the Dataset (probably in the Minerva metadata as a stopgap until Provenance is better worked out)

In this case

  * a query would be run on the source, creating some JSON (or "rows/JSON", or "table/JSON")
  * a Dataset would be created, with metadata pointing back to the Source, and holding the query and query date
  * the result of the query would be stored in a File on the Item
  * it's possible we could store some type or semantics about the results, like whether they should represent points or lines

**Visualization (MapLayer?)**

This one is the most hazy, since we don't have previous work to build on.  The high level definition is *enough info to take an existing Dataset and render it in a Map*.

  * an Item in Girder, living under a User's Minerva/Visualization or Minerva/Map/Visualization folder
  * The metadata would have the mapping params needed to convert the Dataset to something renderable, along with a pointer to the Dataset
  * possibly storing the bytes of the rendering if that made sense (e.g. the GeoJson) in a File, though this sounds more like an implementation detail
  * possibly this could just be some additional object in the Dataset's Minerva metadata, under "visualizations", rather than an Item in its own right

In this case

  * the mapping params from the json to geojson
  * the fact that it outputs geojson
  * possibly the geojson feature type (points or lines)
  * any properties specific to the visualization, e.g.
    * all lines colored red
    * convert all rows to points, with a certain property of that row being used to provide a color, and the color map
    * the params used by clustering display to cluster points

web_external/js/models/MongoSourceModel.js
```diff
@@ -0,0 +1,23 @@
...
+            type: 'POST',
+            data: params,
+            error: null // don't do default error behavior (validation may fail)
+        }).done(_.bind(function (resp) {
+            this.set(resp);
+            this.trigger('m:sourceReceived');
+        }, this)).error(_.bind(function (err) {
+            this.trigger('m:error', err);
+            // If we couldn't successfully start the s3 import destroy the item.
+            this.destroy();
```
mgrauer: You don't need to do this explicit destroy call.  

The S3 Source created an additional folder to store the contents of the files imported from S3, and the destroy call of the Girder Item needed to be overridden so that the additional folder was deleted if there was an error creating the S3 source.

aashish24:
> I'm a bit torn as to whether a Mongo DB or Collection should be a source.

MongoDB should be a source. A source is just a type, an abstract and not an instance. A dataset is an instance, concrete thing so for me Collection should be a dataset.

aashish24:
> a query would be run on the source, creating some JSON (or "rows/JSON", or "table/JSON")

A query should run on collection and hence on a dataset

web_external/js/models/MongoSourceModel.js
mbertrand: Thanks, I removed the destroy call

mbertrand:
@mgrauer, I added two mongo collections to the test containing a couple geojson rows (https://github.com/Kitware/minerva/blob/mongo_source/plugin_tests/data/polygons.json <https://github.com/Kitware/minerva/blob/mongo_source/plugin_tests/data/polygons.json>), with one collection being spatially indexed and one not.  The tests check to make sure that in both cased the geojson is saved in the metadata for datasets created from those collections, but not for the test's original tweets dataset which needs manual intervention to specify latitude and longitude fields.

mbertrand:
Removed the date filter from the widget.

mgrauer:
@mbertrand Thanks, this is pretty close--though I'm not understanding some things.  Your test dataset doesn't have a 2dsphere field, but you create an index on that field in the collection.  

Can you explain what's going on?  Actually, best would be to add some documentation describing what happens in terms of creating and saving geojson in the cases where collections

1. have certain specially named fields
2. have certain specially named indices
3. lack any of these specially named fields or indices

If there are other important considerations please add those.  I'll review this PR in light of the documentation.  You can add it anywhere that makes sense in the docs, or create a new page.  Can you also rebase this branch, and let me know if you need help resolving any conflicts?

mbertrand:
Added some more documentation inside the mong_dataset.py and
mongo_dataset_test.py files.

mbertrand:
@mgrauer, I added some more documentation inside the mongo_dataset.py and
mongo_dataset_test.py files.   '2dsphere' is a type of spatial index on a
field, rather than a field in itself.

The general flow for adding GeoJSON to mongo dataset metadata is:
a) check to see if there is a '2dsphere' index on a field in the
collection.  If true, save the name of the indexed field in the metadata as
'spatial_field'.
b) if there is no '2dsphere' index, look for a 'geometry' field containing
'coordinates' in the sample json row.  If found, save 'geometry' as the
'spatial_field' in the metadada.
c) if there is no 2dsphere index and no 'geometry' field, then don't try to
convert to GeoJSON automatically.  Allow the user to specify latitude and
longitude fields via the Key mapper widget.
d) if a 'spatial_field' key has been assigned in the metadata for the
dataset, use the MongoDataset.convertToGeoJSON method to convert the
collection data to  GeoJSON and save in the metadata.

aashish24:
> The general flow for adding GeoJSON to mongo dataset metadata is:
> a) check to see if there is a '2dsphere' index on a field in the
> collection.  If true, save the name of the indexed field in the metadata as
> 'spatial_field'.
> b) if there is no '2dsphere' index, look for a 'geometry' field containing
> 'coordinates' in the sample json row.  If found, save 'geometry' as the
> 'spatial_field' in the metadada.
> c) if there is no 2dsphere index and no 'geometry' field, then don't try to
> convert to GeoJSON automatically.  Allow the user to specify latitude and
> longitude fields via the Key mapper widget.

@mbertrand the logic makes sense to me.

mgrauer:
>I added some more documentation inside the mongo_dataset.py and
mongo_dataset_test.py files.   '2dsphere' is a type of spatial index on a
field, rather than a field in itself.

That's what I was missing, thanks for the clarification.  Your documentation looks good.  We'll probably want to allow saving the user requested geojson creating by setting a mapping, but we can deal with that in the future when we have a little more clarity on the definitions of our domain objects.

LGTM.

PR #210, opened 2015-11-19
[WIP] Add pip install docs

mgrauer:
A draft based on testing the install path prior to Girder 1.4.0 release.  

Should be updated as part of #204.  I didn't want to lose this info before then.

PR #122, opened 2015-09-24
[WIP] Terra Integration

danlamanna:
@mgrauer Opening this with no intention of merging yet, just as a thread for code review to take place.

PR #211, opened 2015-11-20
Add Postgres source

danlamanna:
@mgrauer No action associated with this source, it also stealth fixes a bug where jade would flip out if your source had no icon/action.

Very simple source, just meant to be used as an input to analyses.

### DIFF
--- a/plugin_tests/postgres_test.py
+++ b/plugin_tests/postgres_test.py
@@ -1,0 +1,90 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+
+###############################################################################
+#  Copyright Kitware Inc.
+#
+#  Licensed under the Apache License, Version 2.0 ( the "License" );
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+###############################################################################
+
+import os
+
+# Need to set the environment variable before importing girder
+os.environ['GIRDER_PORT'] = os.environ.get('GIRDER_TEST_PORT', '20200')  # noqa
+
+from tests import base
+
+
+def setUpModule():
+    """
+    Enable the minerva plugin and start the server.
+    """
+    base.enabledPlugins.append('jobs')
+    base.enabledPlugins.append('romanesco')
+    base.enabledPlugins.append('gravatar')
+    base.enabledPlugins.append('minerva')
+    base.startServer(False)
+
+
+def tearDownModule():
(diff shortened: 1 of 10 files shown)


plugin_tests/postgres_test.py
```diff
@@ -0,0 +1,90 @@
...
+            'name': name,
+            'username': username,
+            'password': password,
+            'dbname': dbname,
+            'baseURL': baseURL
+        }
+        response = self.request(path=path, method='POST', params=params, user=self._user)
+        self.assertStatusOk(response)
+        postgresSource = response.json
+        print(response.json)
```
mgrauer: can you clean this out if it isn't desired?

web_external/templates/body/sourcePanel.jade
```diff
@@ -10,9 +10,11 @@
...
                         - var attributes = {'cid': source.cid}
                         - var sourceType = sourceTypes[source.getSourceType()]
-                        - var classes = 'm-icon-enabled m-display-source-action ' + sourceType.icon + ' ' + sourceType.action
-                        i(class=classes)&attributes(attributes)
-                        //- trash icon to delete source
+
+                        if sourceType
+                          - var classes = 'm-icon-enabled m-display-source-action ' + sourceType.icon + ' ' + sourceType.action
+                          i(class=classes)&attributes(attributes)
+                          //- trash icon to delete source
```
mgrauer: unindent this comment, or better yet, remove it as it is self-evident.

mgrauer:
Is this a WIP, or all you wanted to do here?  It doesn't actually create a postgres connection of any kind.

mgrauer:
>it also stealth fixes a bug where jade would flip out if your source had no icon/action.

the first rule of stealth fixes...

:speak_no_evil:

web_external/js/views/widgets/AddPostgresSourceWidget.js
```diff
@@ -0,0 +1,34 @@
...
+            }, this).createSource(params);
+        }
+    },
+
+    initialize: function (settings) {
+        this.collection = settings.collection;
+    },
+
+    render: function () {
+        var modal = this.$el.html(minerva.templates.addPostgresSourceWidget({}));
```
mgrauer: call `girderModal`, see the note in the dev guide, currently in #210.

danlamanna:
It doesn't create a connection, if you want me to include ```psycopg2``` and setup some basic functionality for executing a query from user input or something, I can. However this is what I had remaining from #122, where a PG source was just used as input to an analysis.

mgrauer:
No need to do that.  I was curious about how much you intended to add with this PR.

danlamanna:
Sure - this is set on my end.

mgrauer:
I made three changes.

1) Rebased on master

2)Reverted the girderModal, b/c this was causing a flash b/w the first dialog (add source) and the next (add postgres source).  I'll need to clarify when to use girderModal in the docs, but I think it is better not to in this case because we have already called girderModal on the first dialog, then the container just gets re-used by the second dialog.

3) Changed the icon to the recently added 'icon-database'.

This branch LGTM if these changes are OK by you.

danlamanna:
image

PR #217, opened 2015-11-20
WMSFeatureInfoWidget minor improvements

mbertrand:
Set maxHeight to 400; prevent an exception that happened occasionally when the value for a feature attribute was null.


mgrauer:
LGTM, merging.

PR #219, opened 2015-11-22
[WIP] Refactor client source model plumbing

mgrauer:
Begins to address #160.

Most client side source models can now be created with 

    minerva.createSourceModel('WmsSourceModel', 'wms', 'minerva_source_wms');

and you won't have to modify the SourceCollection upon adding a new model.  This method's api seem a bit repetitive, you can basically get everything you need from the `source_type`, i.e.

    minerva.createSourceModel('wms');

but I'm erring on the side of greater explicitness for now, and this may prove more helpful once we need to start doing this from other plugins which may not be able to extrapolate the other params from the `source_type`.

I had to add this business with `minerva.clearRegistry('sourceModels');` to start the app in a known good state--when I ran the web_client_test locally it was failing because there were Jasmine timeouts loading certain resources (happened often) and then the test would reload the resources but not create a new application, so the models were throwing an exception because they had already been defined.  This way, whenever the SourceModel.js resource is loaded, it clears out the definitions of any previous SourceModels, and that seems reasonable enough.


danlamanna:
Now might be a good time to discuss the registry in more detail since I'm building a registry (read: object) for  configurable layouts. We need to determine how we're going to store the registry, what type of API we're going to have for interacting with it, and how defensive we want to be.

We should decide if ```createSourceModel``` and ```addPanelGroup``` are bloat, or if they're the only sane way of doing it.

I see the spectrum ranging like this:
- Global object with no restrictions on what gets added, removed, changed, etc.
- Having a model for the client side registry with getters/setters, an API with extremely defensive setters, for example "Added panel objects must have a ```render``` method". This could prove useful for determining when things change, for example "when the map panel view gets replaced with another panel view (like pdf) -> destroy former and render latter".

I personally think we should go with the latter, just curious what your thoughts are.

mgrauer:
@danlamanna Agreed on discussing the registry, that's what I was hoping for here.

I like the more specific and defensive API also, with all of our ideas about extending Minerva through multiple Girder plugins and having configurable and interchangeable layouts/panels, I think the more explicit and defensive we can be the better, it'll save us development and debugging time.

We do have alternatives instead of adding this API, we could `wrap` views and other definitions like in Girder plugins, or we could issue/respond to UI events.  

My inclination is to go the API route because I feel it is the most explicit.  This doesn't prevent developers from doing things like wrapping or redefining, but I think it is an easier path to provide support on.  Could be wrong, and definitely willing to be convinced otherwise.

mgrauer:
If we do use an API, probably best to structure its definition somehow, e.g. putting all methods in `init.js`.  Having parts of it defined (as I did here) in SourceModel.js and parts all over the place would be confusing.

mgrauer:
Closing, the purpose was served.

PR #224, opened 2015-12-01
Remove gaia dependency

mgrauer:


mgrauer:
Removes dependency on Gaia, also cleans up some Travis assumptions that worked only when there was a build cache.

@kotfic  PTAL

kotfic:
@mgrauer I am still having trouble getting the ansible/girder-install to work with minerva on a local integration branch but that isn't a problem with this.

LGTM

Issue #225, opened 2015-12-01
Refactor session model/save


mgrauer:
Right now we save the session in a few ways

  * add/remove a dataset enable the save button
  * save basemap properties through target icon
  * save basemap name/description through session edit
  * save basemap center through session edit

We should unify save methods.

Other changes:

  * currently save session in json file, instead we should use item metadata
  * change session model to extend minervamodel, use the item metadata API in minervamodel

Decide if saving the session saves session data or app state (config/layout) or just as one big blob.

danlamanna:
Additionally, use Backbone get/set. I think (but am not positive) that [this event](https://github.com/Kitware/minerva/blob/8cc22a604396a1f08c7768732503c0406d95ae2b/web_external/js/views/body/SessionView.js#L89-L91) is never executed.

mgrauer:
Be sure to bring back the ability to save the baselayer properties somehow, since they were removed in #226.

mgrauer:
Address [this comment](https://github.com/Kitware/minerva/pull/226/files#r47400183) and [that comment](https://github.com/Kitware/minerva/pull/226/files#r47540677).

mgrauer:
Functionality related to an individual session is

* name
* description
* basemap
* choice of basemap
* zoom level
* center
* currently visualized datasets
* any color mapping/filtering/adapter properties of the visualized dataset
* order of datasets in the stack
* opacity of datasets

mgrauer:
Sessions have been fixed to save in item metadata instead of a json file, in #319.

Issue #186, opened 2015-11-06
specify alternate layout for minerva in dependent plugin


mgrauer:
At least a proof of concept where a Girder plugin dependent on Minerva will specify an alternate layout.

After all the other work in this milestone is done.

Issue #185, opened 2015-11-06
save and load layout configuration


mgrauer:
this can also include the collapsed state of any individual panels

Issue #184, opened 2015-11-06
ability to specify the layout configuration, and specify a default configuration for minerva


mgrauer:
build a system to specify the layout configuration, convert the default layout for minerva to use this system.

PR #226, opened 2015-12-02
Configurable Layout/UI Refactor

danlamanna:
This PR addresses #182, #184, #185, and #186.

The biggest addition is defining a layout for Minerva that can be configured by 3rd party plugins, shown [here](https://github.com/Kitware/minerva/blob/922b531c5c621a37e45e1fee0dcbdda28198e273/web_external/js/views/body/SessionView.js#L120).

Other changes include:
- Removes gridster, also the map slider :facepunch:
- Allowing basemap_args to be stored in the session data, so someone can switch basemaps if they want
- Created a generic Panel view
- Cleans/removes a lot of jade/stylus that seemed extraneous
  - Added some basic mixins to organize how panels are rendered   

Concerns that need to be addressed before merge:
- [x] ~~I'm largely unfamiliar with how this is being used in BSVE, and for Epidemico's work   
     Review from someone with their own projects would be greatly appreciated~~  @mgrauer will make sure this works in the BSVE context before giving final approval
- [x] ~~I removed the baselayer settings button, since I removed the title bar off the map. What is the stopgap to get this merged?~~ noted in #225
- [x] ~~I removed the add dataset button, this should be added back in the form of the + icon similar to sources~~
- [x] The map styling is broken, it goes off the page for some reason, and resizing the window breaks the entire map
- [x] ~~I'm somewhat uneasy about [how panels are added to panel groups](https://github.com/Kitware/minerva/blob/922b531c5c621a37e45e1fee0dcbdda28198e273/web_external/js/views/body/PanelGroup.js#L15-L16) but wasn't entirely sure what the proper Backbone way was~~   
  **Update**: apparently this isn't as hacky as I thought

Future concerns:
- Storing/Loading UI state is somewhat sloppy, there may be better tools to accomodate this type of problem
- Cleaning up the header panel, removing the sessions listing page
- Storing the layout on the session view is sloppy, it should probably be on the model - though it seems like that will be part of a larger refactor in #225


mgrauer:
@danlamanna can you pastry in a screenshot?

danlamanna:
minerva-ui-refactor

mgrauer:
tasty, thanks.

jeffbaumes:
C'mon, you couldn't get a donut logo up there @danlamanna? *disappointed*

jbeezley:
@danlamanna See my changes here: https://github.com/Kitware/minerva/commit/6ede95f70703f14970c74218a501a251310897f9 to fix the map size issues.  I had to go up the div hierarchy to set all the element heights to 100%.  The only other way to so solve this is to explicitly initialize the map with a given size (and manage any and all size changes that may occur).

Note: I also changed how the stylus files are built so that I can get the header height variable inside a different file.  There are other ways to handle this, but importing (or requiring) files from a single place resembles what we should be doing with webpack.

danlamanna:
@jbeezley Thanks for figuring that out, that was more involved than I would have expected.


> resembles what we should be doing with webpack.

:microphone:  :droplet:

jbeezley:
Yeah, I've come to see it like this.  Normally, unless otherwise stated, a parent element is only as large as the content it contains..  Trying to have an element expand to take up 100% of it's parent inverts the usual pattern, so you have to go all the way up the document tree to make it work.  There are all kind of other implications in positioning you run into as well because it disrupts the whole document flow.

web_external/js/views/body/Panel.js
```diff
@@ -0,0 +1,55 @@
...
+
+    getSessionModel: function () {
+        return this.getSessionView().model;
+    },
+
+    /**
+     * Upon confirmation, physically removes the DOM element while
+     * storing the change in the session attributes. This also
+     * enables the save button.
+     **/
```
jbeezley: I take it there is no way to add the panel back after removing it?

danlamanna: Nope, unless you were to manually edit the session.json. It seems like having a dropdown with available panels is something @aashish24 wants at some point - so that could add that functionality.

web_external/js/views/body/SessionView.js
```diff
@@ -93,40 +117,46 @@ minerva.views.SessionView = minerva.View.extend({
...
-            sourceCollection: this.sourceCollection
-        });
-
-        this.sourcePanel = new minerva.views.SourcePanel({
-            session: this.model,
-            sourceCollection: this.sourceCollection,
-            datasetCollection: this.datasetsCollection,
-            parentView: this
-        });
+        this.layout = {
```
jbeezley: Is the layout the kind of thing that could go in the session model?  That would seem more versatile than storing a list of active panels and disabling those that aren't listed.  Then again, I've already demonstrated my lack of understanding of backbone.

danlamanna: From initial message:
> Storing the layout on the session view is sloppy, it should probably be on the model - though it seems like that will be part of a larger refactor in #225

I could do it now, but I feel like a lot of the session is going to be ripped apart shortly.

jbeezley: Fair enough.

jbeezley:
I'm fine with this as is.  It's missing some features, but doesn't back us into any corners that we weren't already camping out in.

docs/extending-minerva.rst
```diff
@@ -0,0 +1,68 @@
+Extending Minerva
+=================
+
+Creating a Minerva Plugin
```
mgrauer: I'm a little confused by this whole section.  Maybe we can define what a "Minerva Plugin" is?

A bit later you say "Minerva specific plugins", so is that different from a "Minerva Plugin"?

I think because some of this stuff is tricky, it is better to err on the side of being overly explicit.


Maybe the title can be "Creating a Minerva plugin through the Girder plugin system"?

and then

"For Minerva specific plugins" => "For Girder plugins intended to extend the functionality of Minerva"?

docs/extending-minerva.rst
```diff
@@ -0,0 +1,68 @@
...
+~~~~~~~~~~~~~~~~~~~~~~~~~
+Minerva utilizes the Girder plugin system, so it will be worthwhile to familiarize yourself with their section on `Plugin Development <http://girder.readthedocs.org/en/latest/plugin-development.html>`_.
+
+For Minerva specific plugins, specifying "minerva" as a dependency within your plugin is required to ensure things load in the proper order. Take this example plugin configuration:
```
mgrauer: "Take this example plugin configuration for a new Girder plugin that extends the functionality of Minerva:"

docs/extending-minerva.rst
```diff
@@ -0,0 +1,68 @@
...
+
+.. code-block:: python
+
+   {
+       "name": "My Minerva Plugin",
+       "dependencies": ["minerva"]
+   }
+
+You should have a Minerva plugin that does nothing, and `can be enabled through the administration console <http://girder.readthedocs.org/en/latest/installation.html#initial-setup>`_.
```
mgrauer: "You should have a Minerva plugin that does nothing" => "Your new Girder plugin should provide no new Minerva functionality at this point"

docs/extending-minerva.rst
```diff
@@ -0,0 +1,68 @@
...
+       "dependencies": ["minerva"]
+   }
+
+You should have a Minerva plugin that does nothing, and `can be enabled through the administration console <http://girder.readthedocs.org/en/latest/installation.html#initial-setup>`_. 
+
+
+Extending the look and feel of Minerva
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+While Minerva plugins follow the guidelines provided by Girder's client side, there are additional concerns when writing for Minerva.
```
mgrauer: Or we can go with "Minerva plugins" as here, we just need to define it more explicitly above.

docs/extending-minerva.rst
```diff
@@ -0,0 +1,68 @@
...
+
+.. note:: Girder has documentation on `extending the client-side application <http://girder.readthedocs.org/en/latest/plugin-development.html#extending-the-client-side-application>`_ which also applies to Minerva.
+
+Minerva Panel Views
+-------------------
+Minerva renders panels in a particular way, as a result there are some guidelines that need be followed when creating your own Panels:
+
+- Panels should never call render within their initialize function
+- Panel views must extend the base Panel view
+- Panel views need to extend their events to take advantage of collapsible, removable, configurable panels
```
mgrauer: this may need to be updated, I'll mention something related in a bit

docs/extending-minerva.rst
```diff
@@ -0,0 +1,68 @@
...
+
+
+Configure Minerva's layout
+--------------------------
+Taking our example plugin from before, we can alter how Minerva displays different panels.
+
+Let's pretend our use case of Minerva deems the Jobs Panel useless, and is more focused on the datasets available. In this case one might want to disable the Jobs Panel entirely, and move the datasets panel to the top.
```
mgrauer: I like the idea of the "fake use case/example", maybe you could set that up a bit more clearly from the start and follow it all the way through.  Could always be done in a later PR though.

docs/extending-minerva.rst
```diff
@@ -0,0 +1,68 @@
...
+
+From your plugin root, create a JavaScript file at ``web_client/js/some-file.js`` and add the following code:
+
+.. code-block:: javascript
+
+   girder.events.once('m:pre-render-panel-groups', function (sessionView) {
+       var leftPanelGroup = _.find(sessionView.layout.panelGroups, function (panelGroup) {
```
mgrauer: This makes me uneasy, doing things directly on the sessionView.layout.panelGroups object.  I'd prefer an API like

    sessionView.getPanelGroup('m-left-panel-group')

even if it is doing the exact same thing under the hood.

docs/extending-minerva.rst
```diff
@@ -0,0 +1,68 @@
...
+           return panelGroup.id === 'm-left-panel-group';
+       });
+
+       // Remove the jobs panel
+       leftPanelGroup.panelViews = _.filter(leftPanelGroup.panelViews, function (panelView) {
```
mgrauer: similarly here, perhaps an API like

    leftPanelGroup.hide('m-jobs-panel')

docs/extending-minerva.rst
```diff
@@ -0,0 +1,68 @@
...
+           return panelView.id !== 'm-jobs-panel';
+       });
+
+       // Move the 'Available Datasets' panel to the top
+       leftPanelGroup.panelViews.sort(function (a, b) {
```
mgrauer: could we pass a sort function to the panelgroup instead?

    leftPanelGroup.sortOrder(function (a, b) ...

plugin_tests/minervaSpec.js
```diff
@@ -11,8 +11,8 @@ describe('a test for the minerva plugin application', function () {
 
     it('displays the minerva app', function () {
         waitsFor(function () {
-            return $('a.m-maps-link:visible').length === 1;
-        }, 'the maps button to be visible');
+            return $('.m-sessions-search-container:visible').length === 1;
+        }, 'the core panel to be visible');
```
mgrauer: maybe "main" or "full-screen" instead of "core"?  

Whatever we go with, since there is a special panel, we should describe that in the docs somewhere.

web_external/js/views/body/MapPanel.js
```diff
@@ -214,9 +215,10 @@ minerva.views.MapPanel = minerva.View.extend({
                     }
                 })
             });
-            this.map.createLayer(this.session.sessionJsonContents.basemap);
+            this.map.createLayer(this.session.sessionJsonContents.basemap,
```
mgrauer: make an API on the session model rather than mucking with the sessionJsonContents directly

docs/extending-minerva.rst
mgrauer: I see in panel.js there is a `removePanel` method.  Should that be used in this example instead of the filter stuff?  or what is the relationship between the different methods of removing the panel?

web_external/js/views/body/AnalysisPanel.js
```diff
@@ -10,7 +10,7 @@
         return name  + 'Widget';
     };
 
-    minerva.views.AnalysisPanel = minerva.View.extend({
+    minerva.views.AnalysisPanel = minerva.views.Panel.extend({
```
mgrauer: Since we are asking panels to extend 'minerva.views.Panel' instead of 'minerva.View' now, can you make the below happen automatically in Panel.js, and then specific panel extensions wouldn't need to do it?

    _.extend(this.events, minerva.views.Panel.prototype.events);

danlamanna: Maybe I'm misunderstanding, but if this were to happen it would have to be in the Panel.js initialize, which would mean extending panels would need to call parent initialize in their initialize. Is that what you're talking about?

docs/extending-minerva.rst
danlamanna: I want to steer clear of the term "hide" because if anyone wants to hide anything they should be using CSS. Everything I'm doing actually *disables* the panel.

The relationship here is that ```removePanel``` is an event which disables a specific panel and then removes the DOM element completely, while this code is hooking in to disable a panel before it is ever rendered.

I will add a method to the session view to disable a given panel and utilize that in removePanel and the documentation.

danlamanna:
@mgrauer I'm going to make some changes and ping you back to take another look, I want to note though that a huge portion of the feedback is directly related to #225, which suggests to me that it may be one of the next things we tackle.

web_external/js/views/body/MapPanel.js
danlamanna: This is essentially what my comment on #225 addresses. I'll change it if you want, but I'm only doing retrieval here and it is identical to how retrieval of session data is done throughout the codebase.

danlamanna:
@mgrauer PTAL. I think I fixed most of your comments, others I responded to in this PR.

mgrauer:
Yeah, definitely punt on anything that will be fixed as part of #225, we should just note it there.

web_external/js/views/body/MapPanel.js
mgrauer: Leave it for that later change.

web_external/js/views/layout/HeaderView.js
```diff
@@ -3,7 +3,7 @@ minerva.views.LayoutHeaderView = minerva.View.extend({
     },
 
     render: function () {
-        this.$el.html(minerva.templates.layoutHeader());
+        //this.$el.html(minerva.templates.layoutHeader());
```
mgrauer: remove this line

web_external/stylesheets/body/mapPanel.styl
```diff
@@ -1,18 +1,10 @@
...
-  margin 0px
-
-  .m-save-current-baselayer
-    float right
-
-  .m-save-current-baselayer:hover
-    color #66a
-    cursor pointer
-
+#map-panel
```
mgrauer: how about 'm-map-panel'?

web_external/templates/body/mapPanel.jade
```diff
@@ -1,5 +1 @@
-.panelTitle
-  | Current Map
-  .m-map-legend-container
-  i.m-save-current-baselayer.icon-target(title="Save current baselayer settings")
 .mapPanelMap
```
mgrauer: maybe 'm-map-panel-map' ?

web_external/js/views/body/SessionView.js
```diff
@@ -93,40 +132,46 @@ minerva.views.SessionView = minerva.View.extend({
...
+            panelGroups: [
+                {
+                    id: 'm-core-panel-group',
```
mgrauer: m-main-panel-group

web_external/stylesheets/body/panelGroups.styl
```diff
@@ -0,0 +1,36 @@
+#m-left-panel-group
+  float left
+  width 280px
+  position absolute
+  z-index 999
+  margin-left 5px
+  top 0
+
+#m-core-panel-group
```
mgrauer: m-main-panel-group

web_external/js/views/body/Panel.js
```diff
@@ -0,0 +1,53 @@
...
+    expandPanel: function (e) {
+        $(e.currentTarget).find('i.icon-down-open').attr('class', 'icon-up-open');
+        this.getSessionModel().addLayoutAttributes(this.el.id, {
+            collapsed: false
+        });
+        this.getSessionView()._enableSave();
+    },
+
+    collapsePanel: function (e) {
+        $(e.currentTarget).find('i.icon-up-open').attr('class', 'icon-down-open');
```
mgrauer: probably beating a :skull: :horse: here, similar to [this comment](https://github.com/Kitware/minerva/pull/226/files#r47400183), what would be nice is if the three methods

    removePanel
    expandPanel
    collapsePanel

would trigger a change on the layout of the session model, and then the session view can be listening for changes on the session model which would re-render the panels and also enable the save button at the appropriate times.

All changes for #225.

web_external/js/views/body/AnalysisPanel.js
mgrauer: You are right, and I think what you've done already is cleanest.  Can you add the line 

    _.extend(this.events, minerva.views.Panel.prototype.events);

to the Extending Minerva docs for great explicitness?

web_external/templates/body/dataPanel.jade
```diff
@@ -1,30 +1,25 @@
...
-                            //- table icon for mapping columns of csv/json
-                            - var displayed = dataset.get('displayed')
-                            if !displayed && (dataset.get('meta').minerva.original_type === 'json' || dataset.get('meta').minerva.original_type === 'mongo') && (!dataset.get('meta').minerva.json_row)
-                                //- when there is no json data, disable the table icon, as if the dataset was displayed
-                                - displayed = true
-                            - var classes = (displayed ? 'icon-table csv-mapping icon-disabled dataset-in-session' : 'icon-table csv-mapping icon-enabled')
-                            i(class=classes)&attributes(attributes)
+include ../layout/panelMixins.jade
 
++panel-title('Available Datasets')
```
mgrauer: Is it possible to put more of this repeated logic in the jade mixin for panel-title?  i.e. passing `#m-data-panel` is the only thing different between here and the jobs panel.  Though maybe this would complicate matters when adding additional icons like in the sources panel?

At the very least, please document this standard jade for a panel in the Extending Minerva doc.

danlamanna:
@mgrauer See [my comments](https://github.com/Kitware/minerva/pull/235#issuecomment-165259049) on #235, aside from that I think we should consider having the panel views call initialize on their parent PanelView and we can take care of the extending of events there, that way if we want to do something else in the future things will "just work".

mgrauer:
Forget what I said on that first PR I created.  I've created a new one #236 to merge into here.  If those are OK with you, you can merge it.

I like your idea of the panel views calling initialize on their parent PanelView.  Can you do that after you've merged #236, and also update the docs? Then I think this would be ready to go.

danlamanna:
Alright @mgrauer - one last look.

mgrauer:
LGTM.

Awesome work @danlamanna .  Thanks for pushing this refactor through, I think it is a big improvement in the UI and in extensibility.

We should be on the lookout for small bugs to clean up in the near future.

mgrauer:
:confetti_ball:

PR #227, opened 2015-12-02
Fix the grunt watch task

### LINKED ISSUE #208: problems with "grunt plugin watch"
(1)

If I run  `grunt plugin watch --verbose` I can see which files are being watched, and it is a number in the thousands, mostly brought in by the [plugin-minerva-build-geojs](https://github.com/Kitware/minerva/blob/c7b82591415c0ee991347ba1b1762e47bf22421e/Gruntfile.js#L195-L198) watch.  This should be a more specific list I'm guessing.

(2)

If I comment out the `watch` section from Minerva's Gruntfile, I see these files watched, so I'm guessing they are set in Girder's plugin watch, but this list doesn't seem desirable.

>Watching plugins/minerva/package.json for changes.
Watching plugins/minerva/.git for changes.
Watching plugins/minerva/analyses for changes.
Watching plugins/minerva/data_services for changes.
Watching plugins/minerva/docs for changes.
Watching plugins/minerva/node_modules for changes.
Watching plugins/minerva/plugin_tests for changes.
Watching plugins/minerva/server for changes.
Watching plugins/minerva/utility for changes.
Watching plugins/minerva/web_client for changes.
Watching plugins/minerva/web_external for changes.

(3)

Our jade, stylus and javascript files are not being watched.

If I change [plugin-minerva-jade-external](https://github.com/Kitware/minerva/blob/c7b82591415c0ee991347ba1b1762e47bf22421e/Gruntfile.js#L183-L186) watch from

    files: _.pluck(grunt.config.get('jade.minerva.files'), 'src'),

to 

    files: ['<%= plugin.minerva.external %>/templates/**/*.jade']

this will correctly watch the jade files.  Unfortunately this now duplicates the listing of the jade files instead of referring to some earlier set variable from the grunt config, and that is much worse in the case of the [uglify task's files](https://github.com/Kitware/minerva/blob/c7b82591415c0ee991347ba1b1762e47bf22421e/Gruntfile.js#L85-L121).

ping @jbeezley for the best way to fix this.

jbeezley:
Since I'm now working on client side code, I found the broken watch task sufficiently annoying that I decided to fix it.

Fixes #208.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -190,31 +190,6 @@ module.exports = function (grunt) {
                 ]
             }
         },
-        watch: {
-            'plugin-minerva-jade-external': {
-                files: _.pluck(grunt.config.get('jade.minerva.files'), 'src'),
-                tasks: ['jade.minerva']
-            },
-            'plugin-minerva-stylus-external': {
-                files: _.pluck(grunt.config.get('stylus.minerva.files'), 'src'),
-                tasks: ['stylus.minerva']
-            },
-            'plugin-minerva-uglify-external': {
-                files: _.pluck(grunt.config.get('uglify.minerva.files'), 'src'),
-                tasks: ['uglify.minerva']
-            },
-            'plugin-minerva-build-geojs': {
-                files: ['<%= plugin.minerva.geojs %>/**/*.js'],
-                tasks: ['shell.minerva-geojs']
-            },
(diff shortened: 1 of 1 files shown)


mgrauer:
Sweet, thanks @jbeezley .

Works for js, jade, and stylus.

PR #228, opened 2015-12-03
[WIP] Local uploader

mgrauer:
This is a big mess, but it worked for an end to end process of

  * upload local pdf to an Item Source
  * display pdf on Item Source
  * Run a pdf table extraction analysis, using Poppler, specifying the pdf item source and page number
  * download the dataset that is a text file, which was output from the extraction job

I'll need to clean this up and see what can be salvaged.

### DIFF
--- a/web_external/js/collections/SourceCollection.js
+++ b/web_external/js/collections/SourceCollection.js
@@ -14,10 +14,12 @@ minerva.collections.SourceCollection = minerva.collections.MinervaCollection.ext
                 return new minerva.models.PostgresSourceModel(attrs, options);
             } else if (attrs.meta.minerva.source_type === 'mongo') {
                 return new minerva.models.MongoSourceModel(attrs, options);
+            } else if (attrs.meta.minerva.source_type === 'item') {
+                return new minerva.models.ItemSourceModel(attrs, options);
             }
         }
 
-        console.error('Source collection includes unknown source type');
+        console.error('Source collection includes unknown source type '+ attrs.meta.minerva.source_type);
         console.error(attrs);
         girder.events.trigger('g:alert', {
             icon: 'cancel',
--- a/analyses/bsve/pdfextraction/minerva.json
+++ b/analyses/bsve/pdfextraction/minerva.json
@@ -1,0 +1,1 @@
+{ "analysis_name": "pdf table extraction", "analysis_type": "pdf_table_extraction" }
--- a/server/jobs/pdf_table_extraction.py
+++ b/server/jobs/pdf_table_extraction.py
@@ -1,0 +1,63 @@
+import json
+import os
+import shutil
+import sys
+import tempfile
+import traceback
+
+from girder.utility import config
+from girder.utility.model_importer import ModelImporter
+from girder.plugins.jobs.constants import JobStatus
+
+import girder_client
+
+
+def run(job):
+    job_model = ModelImporter.model('job', 'jobs')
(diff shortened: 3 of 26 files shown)


mgrauer:
Pasting in some screenshots of functionality introduced here.  We don't need to go with what I did, consider it an exploration.  Let's decide what features from this that we actually want to go into Minerva (or any desired modifications) and I can make a new PR with cleaned up versions.





Here I'm trying to break out functionality as granularly as possible to enable selection and discussion.

  * Added upload icon to sources panel

screen shot 2015-12-03 at 1 46 00 pm


  * Upload icon triggers the upload widget

screen shot 2015-12-03 at 2 10 48 pm

  * Files uploaded through the upload widget become sources (all files uploaded at one time are gathered into a single source) of metadata `source_type: 'item'`
  * sets the metadata `item_type` to the mimetype of the first File
  * there is delay after uploading the bytes of the files, should add some status indicator here
  * shows a document icon for a Source Item in the source panel

screen shot 2015-12-03 at 2 14 32 pm

  * if the item is `item_type` of `application/pdf`, displays a pdf icon instead

screen shot 2015-12-03 at 2 15 59 pm

  * associates a pdf view widget action with the pdf icon, displaying the pdf from the source panel (depends on https://github.com/girder/girder/pull/1123)
screen shot 2015-12-03 at 2 21 18 pm

  * adds a new analysis `pdf table extraction`

screen shot 2015-12-03 at 2 23 14 pm

  * `pdf table extraction` analysis allows you to select a source of `source_type: item` and `item_type: application/pdf`
  * `pdf table extraction` analysis allows you to specify a page to extract.  perhaps better if we can specify a range, or if we specify a single page we have the PDF view jump to that page?  dealing with these PDF plugin viewers isn't easy, and is a bit browser dependent

screen shot 2015-12-03 at 2 23 56 pm

  * running the `pdf table extraction` will run Poppler on the pdf for the given page, extracting text, adding it to a new dataset of `dataset_type: text` which is pretty hacky

  * This branch runs Poppler with Python subprocess (obviously gross), and I've built Poppler locally.  Poppler is GPL, and also I think can be apt-get installed.  Possibly we'd want to break this out into a Romanesco Docker task, but I don't know if there are implications given GPL.  

  * I've added a download icon that downloads the backing Girder Item to datasets of `dataset_type: text` which is just a quick hack for this end-to-end workflow.  I have no idea what we actually want to do with this.  Also depends on the definition of `Dataset`, the thought of which should give everyone the willies.

screen shot 2015-12-03 at 2 30 22 pm

web_external/js/collections/SourceCollection.js
```diff
@@ -14,10 +14,12 @@ minerva.collections.SourceCollection = minerva.collections.MinervaCollection.ext
                 return new minerva.models.PostgresSourceModel(attrs, options);
             } else if (attrs.meta.minerva.source_type === 'mongo') {
                 return new minerva.models.MongoSourceModel(attrs, options);
+            } else if (attrs.meta.minerva.source_type === 'item') {
+                return new minerva.models.ItemSourceModel(attrs, options);
             }
         }
 
-        console.error('Source collection includes unknown source type');
+        console.error('Source collection includes unknown source type '+ attrs.meta.minerva.source_type);
```
danlamanna: +1000

aashish24:
@mgrauer lets talk about this branch next week.

mgrauer:
Closing, superseded by #257.

PR #230, opened 2015-12-03
CSV Viewer 

essamjoubori:
This is the first step to enable the end user to upload csv files and to create a datasets/sources. In this PR the end user will be able to upload csv files and view it in a table format.

screen shot 2015-12-03 at 2 25 29 pm

screen shot 2015-12-03 at 2 32 32 pm

screen shot 2015-12-03 at 2 33 00 pm

screen shot 2015-12-03 at 2 33 13 pm


kotfic:
@essamjoubori Screenshots look really nice,  the one thing that caught my eye is the pager for the CSV viewer.  Is that making a request to the server for new rows when you click "next"  or is the whole table being stored client side?

web_external/js/views/widgets/AddCSVSourceWidget.js
```diff
@@ -0,0 +1,166 @@
...
+        },
+
+        'drop .m-drop-zone': 'filesDropped'
+    },
+
+    parseCsv: function (csv) {
+        var pasredCSV = Papa.parse(csv, {skipEmptyLines: true});
+        if (!pasredCSV || !pasredCSV.data) {
+            console.error('This dataset lacks csv data to create geojson on the client.');
+            return;
```
danlamanna: Is the csv source assumed to be related to geojson?

essamjoubori: Yes, it is related. Fixed the typo thanks!

essamjoubori:
@kotfic No, I'm reading, parsing, and rendering (with [Data Tables](https://datatables.net/)) the whole csv file on the client. Currently, I'm not creating any source/datasets. The next step (still deciding on the right approach) is to save the csv file as a source/datasets and then generate geojson out of it.

mgrauer:
I would like to focus this work solely on displaying tabular data, so let's not do any work related to converting formats or processing the data on this PR.

@essamjoubori A few areas I'd like you to explore and report back on.

  * can the CSV Viewer widget take up most of the screen?  you could add additional classes to the divs with `.modal-dialog` and `.modal-content` to control this
  * can you try using papaparse in combination with the JS File API to only read say 10 rows (or the first $page_size of rows) out of the client side file rather than reading the whole file?
  * similarly, can we get the number of rows out of the file through some combination of the above without reading it all into memory?
  * what is the path to reading only the first $page_size rows out of a csv that is stored server side?  would this work with datatables where we only download the current page (or maybe the current + 1) to the client?
  * does datatables support infinite scrolling in addition to the paging?  what would it take to support this for a client side file and server side file?
  * can we change the display of paging to be first/last/next/prev?
  * slick grid does support infinite scrolling, how much work is it to replace datatables with slick grid, and what are the tradeoffs?
  * what does the search box do?  how would that work with only loading part of the data?
  * is there an option to make the text editable?
  * does the sorting work?  does it need all the rows in memory (probably) ?
  * can we display only certain columns?

essamjoubori:
@mgrauer,

Here are some updates to this PR:

* Showing the total number of rows before viewing the CSV table
* Ability to select all/custom number of rows from the uploaded CSV file to be displayed on the client
screen shot 2015-12-07 at 3 34 20 pm

 
 screen shot 2015-12-07 at 3 37 30 pm

 screen shot 2015-12-07 at 3 38 56 pm

* Ability to load more rows (by clicking on the `Load more rows` button)
* Hide/show columns

 screen shot 2015-12-07 at 3 41 10 pm


Regarding other points: 


* what is the path to reading only the first $page_size rows out of a csv that is stored server side? would this work with datatables where we only download the current page (or maybe the current + 1) to the client?
  Datatables data source can be obtained via Ajax calls. We can use papaparse on the server to parse the n requested rows and to render them on the client

* does datatables support infinite scrolling in addition to the paging? what would it take to support this for a client side file and server side file?
This feature was removed in the new version

* slick grid does support infinite scrolling, how much work is it to replace datatables with slick grid, and what are the tradeoffs?
The only worry I have is that Slick grid project is not well maintained (274 issues and 138 PRs) and might create issues for us in the future

* is there an option to make the text editable?
There is an extension to the library that enables you to edit the data; however it's not free. Making the text editable is doable with jQuery and Ajax calls

* does the sorting work? does it need all the rows in memory (probably) ?
Yes. It will sort the viewed data (only the parsed rows)

* what does the search box do? how would that work with only loading part of the data?
It searches the table (text matching) and will only work with the parsed rows

mgrauer:
Awesome, thanks for the updates.  I may ask you to give a demo at the 330 meeting today, or maybe stay on after to go over it in a bit more detail.

aashish24:
@essamjoubori do you have any more changes pending?

essamjoubori:
@aashish24 @mgrauer I still need to add a few features before it's ready for QA.

essamjoubori:
This PR is now PR#283 as the latter was built on top of PR#257

Issue #204, opened 2015-11-17
Peg Minerva travis to next Girder minor release


mgrauer:
probably 1.4.0.

PR #231, opened 2015-12-07
Set girder to 1.4.1 for travis

mgrauer:
Spark was having some issues on #226 , but also it looks like the mean contour analysis was running into a breakage in a key not being returned by Girder.

This PR attempts to run travis tests pegging Girder to v1.4.1.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,7 +28,7 @@ before_install:
     - cd $HOME/build
     # remove existing girder dir, it remains from caching operations
     - rm -rf girder
-    - git clone https://github.com/girder/girder.git
+    - git clone https://github.com/girder/girder.git --branch v1.4.1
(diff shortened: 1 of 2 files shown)


mgrauer:
@jbeezley @danlamanna PTAL

This fixes the breaking mean_contour_analysis test.  Spark was having issues in some Travis runs (not sure why, I cleared the Travis cache in #226 and that seemed to fix the Spark master problems), but even when Spark worked, our test depends on a [key that has been removed](https://github.com/girder/girder/pull/1122#issuecomment-162708824).  

I've added another commit that removes our dependence on the `sha512` key so that this test will work when we bump the Girder version in Minerva's Travis past 1.4.1.

Fixes #204.

jbeezley:
LGTM.  All changes are breaking changes.

PR #232, opened 2015-12-08
Remove cdn client resources

mgrauer:
At first this was intended to fix up the cdn resources that were hard-coded to http.  @jbeezley pointed out how this was a problem when Minerva was served from https, the browser would block the mixed content.  PTAL.

In looking at those resources, I realized we aren't using any of them anyway and they can be removed.

@essamjoubori When we decide on a table library we want to bring in, we can add it in then.  Datatables is packaged for bower and npm, so that should be easy if we go with it.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -283,9 +283,6 @@ module.exports = function (grunt) {
             }
         }
         dependencies.concat([
-            'http://cdn.datatables.net/1.10.7/js/jquery.dataTables.min.js',
-            'http://cdn.jsdelivr.net/momentjs/2.9.0/moment.min.js',
-            'http://cdn.jsdelivr.net/bootstrap.daterangepicker/1/daterangepicker.js',
             '/' + staticDir + '/papaparse.min.js',
             '/' + staticDir + '/jsonpath.min.js'
         ]);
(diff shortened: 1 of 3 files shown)


danlamanna:
Aren't they being included via [the Gruntfile](https://github.com/Kitware/minerva/blob/7b146d4ce8a13e655d9c85f9209a8042d694795a/Gruntfile.js#L286-L288) anyway?

Additionally, what happens to the [KeyMapWidget](https://github.com/Kitware/minerva/blob/7b146d4ce8a13e655d9c85f9209a8042d694795a/web_external/js/views/widgets/KeymapWidget.js#L113)?

mgrauer:
Arg.  Now I'm even more determined to remove them.  Thanks for noticing.  

I'll remove that Date filter stuff from the KeyMapWidget, it isn't used by the jade anyway.  We wanted to take it out before, just didn't remove it all the way.

jbeezley:
I think the resources are in the Gruntfile so that they get built into the `testEnv.html`.

mgrauer:
@danlamanna I cleared out the cdn and what uses them.  PTAL.

danlamanna:
LGTM

PR #233, opened 2015-12-11
Add numpy to the list of python dependencies

jbeezley:
We use numpy explicitly in a couple of places, and it is no longer installed via `gaia`.


danlamanna:
LGTM

PR #234, opened 2015-12-11
Added a BSVE CLI to help work with their API

jbeezley:
I decided to cherry pick my other BSVE branch to push in the server side content alone without the client side components, which will take longer to finish.  I still need to fix the analysis test, but I think the actual search functionality still works.


jbeezley:
Update: fixed the test.  This is ready for review.

mgrauer:
LGTM.  

It might be nice to have the CLI take a path to a JSON file of the credentials or pull from ENV vars, but I won't request these features until I have an actual need.

danlamanna:
No [click](http://click.pocoo.org/5/)? :P

jbeezley:
The CLI does actually use [environment variables](https://github.com/Kitware/minerva/blob/bsve-search-utility/server/utility/bsve/bsve_utility.py#L329-L340): `BSVE_USERNAME`, `BSVE_APIKEY`, `BSVE_SECRETKEY`.  I set those in bashrc so I don't have to worry about it.

jbeezley:
@danlamanna I actually found click after I had done most of this, otherwise I would have considered it.

PR #240, opened 2015-12-29
Minerva vagrant

kotfic:
Adds vagrant file and ansible support for installing Minerva on a virtual box.  To test (assuming vagrant is installed and working)  checkout this branch and run ```vagrant up```

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -5,6 +5,8 @@ Vagrant.configure(2) do |config|
   config.vm.box = "ubuntu/trusty64"
   host_port = ENV["GIRDER_HOST_PORT"] || 8080
 
+  sync_folders = ENV["DEVELOPMENT_SYNC_FOLDERS"] || false
+
   config.vm.network "forwarded_port", guest: 8080, host: host_port
 
   config.vm.define "minerva" do |node| end
@@ -37,17 +39,23 @@ Vagrant.configure(2) do |config|
   end
 
 
-  if File.directory?("../girder")
-    config.vm.synced_folder "../girder", "/opt/girder", owner: 1003, group: 1003
-  end
-
-  if File.directory?("../romanesco")
-    config.vm.synced_folder "../romanesco", "/opt/romanesco", owner: 1002, group: 1002
-  end
-
-
   config.vm.synced_folder ".", "/vagrant", disabled: true
-  config.vm.synced_folder ".", "/opt/minerva", owner: 1003, group: 1003
+
+  if sync_folders
+    if File.directory?("../girder")
+      config.vm.synced_folder "../girder", "/opt/girder", owner: 1003, group: 1003
+    end
+
+    if File.directory?("../romanesco")
+      config.vm.synced_folder "../romanesco", "/opt/romanesco", owner: 1002, group: 1002
+    end
+
+    config.vm.synced_folder ".", "/opt/minerva", owner: 1003, group: 1003
+  end
+
+
+
+
 
   config.vm.provision "ansible" do |ansible|
     ansible.groups = {
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -1,0 +1,64 @@
+# -*- mode: ruby -*-
+# vi: set ft=ruby :
+
+Vagrant.configure(2) do |config|
+  config.vm.box = "ubuntu/trusty64"
+
+  config.vm.network "forwarded_port", guest: 8080, host: 8080
+
+  config.vm.define "minerva" do |node| end
+
+  config.vm.provider "virtualbox" do |vb|
+    host = RbConfig::CONFIG['host_os']
+
+    # Give VM 1/4 system memory & access to 1/2 of the cpu cores on the host
+    if host =~ /darwin/
+      cpus = `sysctl -n hw.ncpu`.to_i
+      cpus = cpus / 2
+      # sysctl returns Bytes and we need to convert to MB
+      mem = `sysctl -n hw.memsize`.to_i / 1024 / 1024 / 4
+    elsif host =~ /linux/
+      cpus = `nproc`.to_i
+      cpus = cpus / 2
+      # meminfo shows KB and we need to convert to MB
+      mem = `grep 'MemTotal' /proc/meminfo | sed -e 's/MemTotal://' -e 's/ kB//'`.to_i / 1024 / 4
+    else # Guess!
+      cpus = 2
+      mem = 4096
+    end
+
+    vb.customize ["modifyvm", :id, "--memory", mem]
+    vb.customize ["modifyvm", :id, "--cpus", cpus]
+    vb.customize ["modifyvm", :id, "--nictype1", "virtio"]
+    vb.customize ["modifyvm", :id, "--natdnshostresolver1", "on"]
+    vb.customize ["modifyvm", :id, "--natdnsproxy1", "on"]
+
+  end
+
+
+  if File.directory?("../girder")
+    config.vm.synced_folder "../girder", "/opt/girder", owner: 1003, group: 1003
+  end
+  if File.directory?("../romanesco")
+    config.vm.synced_folder "../romanesco", "/opt/plugins/romanesco", owner: 1002, group: 1002
+  end
+
+
+  config.vm.synced_folder ".", "/vagrant", disabled: true
+  config.vm.synced_folder ".", "/opt/plugins/minerva", owner: 1003, group: 1003
+
+  config.vm.provision "ansible" do |ansible|
+    ansible.groups = {
+      "all" => ['minerva'],
+      "girder" => ['minerva'],
+      "mongo" => ['minerva'],
+      "rabbitmq" => ['minerva']
+    }
+
+    ansible.extra_vars = {
+      default_user: "vagrant"
+    }
+
+    ansible.playbook = "ansible/site.yml"
+  end
+end
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -3,8 +3,9 @@
 
 Vagrant.configure(2) do |config|
   config.vm.box = "ubuntu/trusty64"
+  host_port = ENV["GIRDER_HOST_PORT"] || 8080
 
-  config.vm.network "forwarded_port", guest: 8080, host: 8080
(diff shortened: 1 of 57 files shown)


ansible/roles/girder/meta/main.yml
```diff
@@ -0,0 +1,8 @@
+---
+
+dependencies:
+  - role: girder-variables
+  - role: girder-install
+#  - role: girder-grunt
```
mgrauer: cleanup

ansible/roles/romanesco-install/templates/run-service.j2
```diff
@@ -0,0 +1,14 @@
+#! /usr/bin/env bash
+
+# f="$( ls /opt/uvcdat/*/bin/setup_runtime.sh 2>/dev/null | tail -n 1 )"
+# if [ -n "$f" ] ; then source "$f" ; fi
```
mgrauer: can we get rid of these two lines?

ansible/roles/upstart/meta/main.yml
```diff
@@ -0,0 +1,9 @@
+---
+
+dependencies:
+  - role: user-generate
+    name: "{{ user }}"
+    # group: "{{ group }}"
```
mgrauer: cleanup

mgrauer:
This is fantastic @kotfic .  It works and I give it a LGTM.

We may want to revert the geojs SHA to a specific commit, but I'll probably follow up soon with a PR doing that anyway.

Here's my wishlist of improvements for this ansible work, though I do not want them on this branch.  I'm only putting there here for discussion.  After you have a chance to respond, I can make issues out of them.


  * add documentation to use Vagrant/ansible, we could probably adapt from [Covalic](https://github.com/girder/covalic/blob/master/vagrant.md)
  * add the ability to specify an install location for girder through a variable
  * add the ability specify an assetstore file path through a variable
  * I'd prefer not syncing the Minerva (and Girder, but less so) folders with the VM, for two reasons.  One, I've done work on a VM before inside git, and forgot to push to a remote, then brought the VM down and back up.  The folder syncing on VirtualBox then syncs from the host to the VM, destroying whatever was in the VM for those synced folders.  Two, it would allow to specify a different version of Minerva to install onto the VM than what is on the host.
  * related to the above, add the ability to specify a Minerva version
  * add the ability to update the version after initial provisioning, and rebuild/restart whatever is necessary for the below projects.  As an example, this is done in Covalic with the [`deploy-update` tag](https://github.com/girder/covalic/blob/8aae649c30ec165d6feb57becf161f492fcb98dd/ansible/roles/girder/tasks/main.yml#L52), though it's possible that (Covalic and Minerva)'s ansible code may unify further and so that example could change.
      * romanesco
      * girder
      * minerva
  * ability to specify and install Minerva analyses (I know this depends on some work in [Girder](https://github.com/girder/girder/pull/1151))
  * ability to install girder (really to call `girder-install`) as the girder user without using sudo--this may require using a virtual env.  
  * ability to set a different service script for Girder
  * addition of apache on the VM that proxies 80 to CherryPy running on 8080

mgrauer:
Also, if this is to be used for developers, there should be a recommended quickest path to build Girder or Minerva code that has changed.

kotfic:
@mgrauer In terms of folder syncing,  what we've done on HPCCloud is provide a DEVELOPMENT environment variable,   if DEVELOPMENT=1  then we sync folders,  otherwise it checks out from a branch or hash. I can address this and a few other things in this branch,  some of the other things (e.g., adding apache,  analyses etc) will need more discussion.

kotfic:
@mgrauer PTAL -  this installs girder and everything else into a virtualenv at /opt/.girderenv.  As a bonus I've tested deploying to an AWS instance and it passes ad-hoc smoke tests.

mgrauer:
Hurm.  I'm guessing this is a temporary thing, I'll try again.

TASK: [minerva | Install minerva system dependencies] *************************
fatal: [minerva] => Traceback (most recent call last):
  File "/Users/mgrauer/dev/geo/ansible/lib/ansible/runner/__init__.py", line 586, in _executor
    exec_rc = self._executor_internal(host, new_stdin)
  File "/Users/mgrauer/dev/geo/ansible/lib/ansible/runner/__init__.py", line 789, in _executor_internal
    return self._executor_internal_inner(host, self.module_name, self.module_args, inject, port, complex_args=complex_args)
  File "/Users/mgrauer/dev/geo/ansible/lib/ansible/runner/__init__.py", line 983, in _executor_internal_inner
    except errors.AnsibleConnectionFailed, e:
AttributeError: 'module' object has no attribute 'AnsibleConnectionFailed'


FATAL: all hosts have already failed -- aborting

PLAY RECAP ********************************************************************
           to retry, use: --limit @/Users/mgrauer/site.retry

minerva                    : ok=113  changed=54   unreachable=1    failed=0

Ansible failed to complete successfully. Any error output should be
visible above. Please fix these errors and try again.

mgrauer:
I've now repeated this error at the same step, with ansible version 1.9.3, after deleting the previous VM and .vagrant dir. Have you seen this before @kotfic ?  Definitely could be an artifact of my environment somehow.

danlamanna:
FWIW - `vagrant up` successfully ran for me on the first try (running debian with ansible 1.9.4).

Vagrantfile
```diff
@@ -0,0 +1,66 @@
...
+    vb.customize ["modifyvm", :id, "--cpus", cpus]
+    vb.customize ["modifyvm", :id, "--nictype1", "virtio"]
+    vb.customize ["modifyvm", :id, "--natdnshostresolver1", "on"]
+    vb.customize ["modifyvm", :id, "--natdnsproxy1", "on"]
+
+  end
+
+
+  if File.directory?("../girder")
+    config.vm.synced_folder "../girder", "/opt/girder", owner: 1003, group: 1003
```
mgrauer: what are these group numbers 1003 and 1002?

mgrauer:
I figured out what's going on.  My host is a mac, maybe that is the difference with what @danlamanna  sees?

The last task that works is the `- name: minerva | owner | set`, which changes ownership and perms on everything under minerva on the VM, which includes the `.vagrant` dir, so that is what breaks the very next ansible task on the host due to the synced files.

Some combination of vagrant/ansible/mac balks at the inventory file being executable and the following task fails.  I also can't run `vagrant provision` at this point, but as soon as I run this

    chmod -x .vagrant/provisioners/ansible/inventory/vagrant_ansible_inventory

I can run `vagrant provision` again.

I removed all of the file syncing from the Vagrant file on my host and provisioning appeared to finish correctly, from start to finish.  

Is the file syncing important to any of your workflows?  To me it is an impediment (beyond this immediate problem), so I don't mind if they go.  If they stay, maybe the minerva ownership could more specifically target or exclude the `.vagrant` dir.  I also have a question about the group numbers in the file syncing--that needs to be commented/explained.

I also don't want to lose track of the fact that these are pretty sweet improvements @kotfic.

:sun_with_face:

kotfic:
@mgrauer the 1002, 1003 buisness has to do with vagrant syncing folders before the girder and romanesco users are created. Changing the user/group of a sync'ed folder on the VM is difficult (impossible?) after it has been synced so we set the uid/gid of the users that will be created by the ansible script.

Clearly an intolerably dirty hack.

I agree that folder syncing is not strictly necessary.  I'm willing to drop it in light of the above issues assuming it doesn't drastically effect anyone else's workflow.

mgrauer:
We've agreed to

1) remove all filesyncing for the time being from the Vagrantfile
2) peg girder_version to 1.4.1
3) peg romanesco_version to cfdc285fd2d94

and that should be enough to get this merged

aashish24:
> 1) remove all filesyncing for the time being from the Vagrantfile
> 2) peg girder_version to 1.4.1
> 3) peg romanesco_version to cfdc285fd2d94

All sounds good to me :+1:

kotfic:
@mgrauer @danlamanna this is ready for review

Vagrantfile
```diff
@@ -0,0 +1,74 @@
...
+      "girder" => ['minerva'],
+      "mongo" => ['minerva'],
+      "rabbitmq" => ['minerva']
+    }
+
+    ansible.extra_vars = {
+      default_user: "vagrant"
+    }
+
+    ansible.playbook = "ansible/site.yml"
```
danlamanna: Do these paths fall victim to the same issue as Kitware/gobig#48?

danlamanna:
@kotfic Did the merge overwrite the `romanesco_version` pinning? I still see it as master.

ansible/site.yml
```diff
@@ -0,0 +1,120 @@
...
+      file:
+        path: "{{venv_root}}"
+        state: directory
+        mode: 0775
+      sudo: yes
+
+  roles:
+    - role: romanesco
+      rabbitmq_ansible_group: rabbitmq
+      romaneco_version: "cfdc285fd2d94"
```
danlamanna: Here ^^

Vagrantfile
```diff
@@ -0,0 +1,74 @@
...
+    vb.customize ["modifyvm", :id, "--natdnsproxy1", "on"]
+
+  end
+
+
+  config.vm.synced_folder ".", "/vagrant", disabled: true
+
+  if sync_folders
+    if File.directory?("../girder")
+      config.vm.synced_folder "../girder", "/opt/girder", owner: 1003, group: 1003
```
danlamanna: Can synced_folders take strings of the user/group rather than the magic numbers? Or a comment indicating what they are, I'm assuming vagrant/vagrant?

Vagrantfile
kotfic: Not directly. The ```config.vm.provision``` section is only run on ```vagrant up``` or ```vagrant provision```  which is run from the same directory as the vagrant file.

mgrauer:
After running this I have a working Minerva, and the romanesco/celery worker is running, as is rabbit, but Spark is not running.

If that makes sense to you @kotfic, LGTM.

kotfic:
@mgrauer yeah, i  think spark is out of scope for this - ideally cumulus is going to be responsible for deploying spark from the minerva app. l I am going to merge.

PR #241, opened 2015-12-30
Proxy non-auth wms

mgrauer:
Ready for review.  Maybe @mbertrand  or @jbeezley can take a look?

This proxies non authenticated wms layers through Girder as well as authenticated layers.  This change allows us to control the scheme wms resources are served through, so that we don't serve `http` wms resources that are out of our control when Minerva is served from `https`, since the browser won't display the mixed `http` content in this case.

### DIFF
--- a/server/loader.py
+++ b/server/loader.py
@@ -35,10 +35,15 @@ from girder.plugins.minerva.utility.minerva_utility import decryptCredentials
 class WmsProxy(object):
     exposed = True
 
-    def GET(self, url, credentials, **params):
-        auth = 'Basic ' + b64encode(decryptCredentials(bytes(credentials)))
-        headers = {'Authorization': auth}
-        r = requests.get(url, params=params, headers=headers)
+    def GET(self, url, **params):
+        if 'minerva_credentials' in params:
+            creds = params['minerva_credentials']
+            del params['minerva_credentials']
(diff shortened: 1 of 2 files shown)


aashish24:
> so that we don't serve http wms resources that are out of our control when Minerva is served from https

@mgrauer so if we have a wms server serving on http, minerva would not serve it and  present an error or some sort?

jbeezley:
Image tiles will load but with copious warnings.  Other resources fetched via http will fail to load.

aashish24:
> Image tiles will load but with copious warnings. Other resources fetched via http will fail to load.

thanks @jbeezley I question was what Minerva is going to do in such cases? I don't see any error / warning handling so is it upto the user to look at the console error messages?

jbeezley:
@aashish24 I think the point of this PR is to avoid these errors by proxying them through cherrypy.

I do have concerns about this in general that is not specific to this PR.  I don't think this approach to proxying will be scalable to a deployable instance.  To proxy static content like this for real, we will need to go through nginx.

That said, this PR does at least solve a problem with our current deployment, so I'll give it a +1 to merge.

mbertrand:
LGTM.

aashish24:
Thanks @jbeezley makes sense to me. LGTM as well.

PR #250, opened 2016-01-11
MMWR data import

jbeezley:
This implements a new analysis to download and accumulate MMWR by state from the BSVE.  Note, this dataset is in it's entirety [here](https://data.cdc.gov/resource/qpap-3u8w.json), but this accesses it through the BSVE endpoint at `/api/data/query/SODA`.  Currently if you try to display the data, it will appear as a solid red outline because the polygon interiors are not styled.  I will add display options for the choropleth in a separate PR.


aashish24:
Nice @jbeezley :bell:

jbeezley:
The `web_client_minerva` test is failing on travis, but not for me locally.  I'm rerunning the test on master as I suspect some external factor broke it.

jbeezley:
#250 resolves the test failure here.

server/jobs/soda_import_worker.py
```diff
@@ -0,0 +1,268 @@
...
+                    print('Encountered a negative value!')
+
+                # accumulate the field as a property
+                props[key] = props.get(key, 0) + v
+
+                # add the property to the global list of properties found
+                all_props.add(key)
+
+    # filter out states that have no records
+    # geojson['properties'] = filter(
```
mgrauer: do you want this commented out section or should it be cleaned up?

jbeezley: I wasn't sure at the time.  Probably should remove it.

analyses/bsve/mmwr/minerva.json
```diff
@@ -0,0 +1 @@
+{ "analysis_name": "MMWR data import", "analysis_type": "mmwr_data_import" }
```
mgrauer: This needs to be changed to `mmwr_import_data` to match up with the widget name.

mgrauer:
Nice @jbeezley.

I needed to rename the analysis, or otherwise the correct widget object wasn't found.

Other than that, LGTM.

I noticed that the interaction with the map within Minerva is less than ideal--the US was displayed partially hidden under the control panels, and I couldn't "grab and translate" the map to where the US was centered, so I had to zoom out to where the map image was a much smaller portion of the map container.  I'll be on the lookout for changes here once we update to GeoJs 0.6.0, as that might be different.  Zach also advised setting opacity on the control panels to 0.85, then transitioning to opacity 1 over 400 ms upon hover, which would help with this display hidden by controls, so that should be an easy improvement.



screen shot 2016-01-12 at 12 54 09 pm

jbeezley:
I fixed the analysis name.  I kept changing my mind on how to name things.

There are a lot of improvements that can and should be made to the map.  Version 0.6 will bring us a lot more options for configuration and testing.  I'm hoping we can make a push on this in the near future.

mgrauer:
> I kept changing my mind on how to name things.

I try to limit this to 50% of my brain cycles.  Have yet to be successful.

LGTM.

PR #251, opened 2016-01-11
Downgrade romanesco on travis for girder compatibility

jbeezley:
This will (hopefully) fix the failing client test.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,6 +35,8 @@ before_install:
     # clone romanesco
     - cd $HOME/build/girder/plugins
     - git clone https://github.com/Kitware/romanesco.git
+    - cd romanesco
+    - git checkout cfdc285fd2d94ac1b8793b33e5ef3b49e59ccbb3 # after this, need to update girder
     - cd $HOME/build
 
     - cd girder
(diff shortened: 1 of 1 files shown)


aashish24:
@jbeezley does this SHA works with girder master as of today?

jbeezley:
Probably, but I'm not sure.  The point is that it *does* work with the version on girder we install on Travis.  I could have updated girder but there is a much greater chance that would break something else in the process.

aashish24:
> could have updated girder but there is a much greater chance that would break something else in the process.

Sounds reasonable. Thanks,

mgrauer:
@jbeezley This looks fine, thanks for fixing, but I don't understand your comment.  Can you provide more context?

    # after this, need to update girder

jbeezley:
If we update romanesco passed that SHA, we need to update girder as well.

mgrauer:
Thanks @jbeezley, I added that clarification to the branch.  I was reading it "after this" as something I needed to do further down in the Travis file.

LGTM.

jbeezley:
Ah yes, that's much better.

Issue #157, opened 2015-10-19
Dataset models should determine how they are added to the map panel


danlamanna:
Right now MapPanel.js is an area where each individual dataset needs to define how it gets rendered within one if statement. This should be generalized in some manner, maybe an individual dataset model can specify it's own ```addDataset``` which will handle the rendering.

aashish24:
@danlamanna I am not sure if having a addDataSet int he model itself would be the best thing. Some alternate solution would be:

1. have some sort of mapper objects that can take the data and the map?
2. have some sort of layer util which can generate new layers depending on the dataset type?

mgrauer:
Yes, this is a constant sore spot.  You can see that I went [back](https://github.com/Kitware/minerva/commit/2dca2b2967cd354c64e899640f375d5daaa925e9) and [forth](https://github.com/Kitware/minerva/commit/3ce632fa405d2de7f8b7b56f20a7301263012ae7) on this very issue when we built support for WMS layers in #115.

I don't like the idea of defining it in the model, since that seems to violate MVC separation, and is the reason I moved the WMS rendering back to the MapPanel.

Having the mapper/layer util is probably the best path forward.  Some dataset types like contour are lucky enough that they can just provide a reader.  Geojson works like this in the general case, but if you want to do anything custom to the rendering, this [isn't so straightforward](https://github.com/Kitware/minerva/pull/143#issuecomment-146874023).

It might be nice to wait until we have one or two more dataset types before refactoring this to get a better sense of the needs.

Related issues

  * a dataset type will have to provide a way to remove itself from the map
  * if/when we split out the legend from the map, datasets will need to specify how to add/remove from the legend
  * any new dataset type will have to define a reader/mapper
  * we will need to define a similar and related path for datasets to display on time serious plots
  * where do we store all of the properties that specify how we want the dataset rendered (e.g. color tied to a particular field, or opacity)?  they probably don't belong on the dataset itself, maybe in some object tied to the dataset and stored in the session?

@danlamanna if you have a dataset type not already in master, please report on the relevant details for how it is working for you.  Also I'm curious to get your take on this w.r.t. time series plots.

mgrauer:
After looking at #205, I have some new thoughts for direction here.

1. Introduce Layer or LayerView objects that understand how different map layers work, and can have the appropriate layer type created for a dataset.  These are Rendering Adapters that are specific to the Dataset type and also to the fact that the Dataset is being rendered in a GeoJs map.
[...]

PR #256, opened 2016-01-13
Client side handling of MMWR data

### LINKED ISSUE #244: Summarize and visualize MMWR data in GeoJS/Minerva

### LINKED ISSUE #245: Provide interactive pop-ups on MMWR dataset visualization

### LINKED ISSUE #246: Provide mapping options (visualization) for MMWR dataset
This would mean that user can chose a color scheme etc. This might be a much bigger task.

jbeezley:
This is the initial work for the MMWR data display.  It is pretty basic.  There is a widget to set the value used to color the choropleth generating by attributes in the dataset
screen shot 2016-01-13 at 1 37 33 pm

And, you can render it like so:
screen shot 2016-01-13 at 1 38 16 pm

The code here is specialized for only this data source because making it general can only reasonably be done as part of a more ambitious refactoring of the dataset models.  There are a number of directions I can go from here:
* Configuration of the color scheme: Like what @mgrauer did in https://github.com/Kitware/minerva/pull/143
* Popup information on click: Like what @mbertrand did in https://github.com/Kitware/minerva/pull/166
* Performance: the performance of the polygon rendering is pretty terrible.  There are number of offenders, the worst seems to be triangulating the polygons, which takes about 30 seconds.  I tried a smaller dataset for the state borders, but it didn't help much and the quality of the polygons was noticeably worse.

Closes #244 
Closes #245 
Closes #246

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -76,7 +76,10 @@ module.exports = function (grunt) {
         stylus: {
             minerva: {
                 files: [{
-                    src: ['<%= plugin.minerva.external %>/stylesheets/main.styl'],
+                    src: [
+                        '<%= plugin.minerva.external %>/stylesheets/main.styl',
+                        '<%= plugin.minerva.root %>/node_modules/colorbrewer/colorbrewer.css'
+                    ],
                     dest: '<%= plugin.minerva.static %>/minerva.min.css'
                 }]
             }
@@ -100,7 +103,8 @@ module.exports = function (grunt) {
                             '<%= plugin.minerva.source %>/models/**/*.js',
                             '<%= plugin.minerva.source %>/collections/**/*.js',
                             '<%= plugin.minerva.source %>/views/body/Panel.js',
-                            '<%= plugin.minerva.source %>/views/**/*.js'
+                            '<%= plugin.minerva.source %>/views/**/*.js',
+                            '<%= plugin.minerva.root %>/node_modules/colorbrewer/colorbrewer.js'
                         ],
                         dest: '<%= plugin.minerva.static %>/minerva.min.js'
                     },
--- a/server/jobs/soda_import_worker.py
+++ b/server/jobs/soda_import_worker.py
@@ -25,8 +25,8 @@ import girder_client
 
 #: geojson of state boundaries source url
 states_source = (
-    'https://gist.githubusercontent.com/jbeezley/82150e34cee8512815cf/'
-    'raw/32549660c18149698a849b00cac6df8ddd2d1f1a/us-states.json'
+    'https://gist.githubusercontent.com/jbeezley/ec21c4a016a84c5def74/'
+    'raw/0eb947eccbba98c01caad6ac7fbc7f8fe8334231/us-states.geojson'
 )
 
 #: cached geojson object
--- a/server/jobs/soda_import_worker.py
+++ b/server/jobs/soda_import_worker.py
@@ -234,6 +234,10 @@ def run(job):
         minerva_metadata['dataset_type'] = 'geojson'
         minerva_metadata['values'] = data['properties']['values']
 
+        # set a default "color-by" attribute if possible
+        if data['properties']['values']:
+            minerva_metadata['colorByValue'] = data['properties']['values'][0]
+
         mM(dataset, minerva_metadata)
 
         existing = file_model.findOne({
--- a/server/jobs/soda_import_worker.py
+++ b/server/jobs/soda_import_worker.py
@@ -106,12 +106,15 @@ def accumulate(data):
         _states = requests.get(states_source).json()
     geojson = copy.deepcopy(_states)
 
-    # create a abbr -> properties mapping
-    states = {
-        feature.get('properties', {}).get('abbr'):
-            feature.get('properties', {})
-        for feature in geojson.get('features', [])
-    }
+    # create a abbr -> properties mappin and remove unwanted properties
+    states = {}
+    for feature in geojson.get('features', []):
+        abbr = feature['properties']['abbr']
+        name = feature['properties']['name']
+        feature['properties'].clear()
+        feature['properties']['abbr'] = abbr
+        feature['properties']['name'] = name
+        states[abbr] = feature['properties']
 
     # city, state regex
     citystate = re.compile(r', ([A-Z][A-Z])$')
@@ -237,6 +240,7 @@ def run(job):
         # set a default "color-by" attribute if possible
         if data['properties']['values']:
             minerva_metadata['colorByValue'] = data['properties']['values'][0]
+        minerva_metadata['colorScheme'] = 'YlOrRd'
 
         mM(dataset, minerva_metadata)
 
--- a/web_external/js/views/body/MapPanel.js
+++ b/web_external/js/views/body/MapPanel.js
@@ -137,7 +137,7 @@ minerva.views.MapPanel = minerva.views.Panel.extend({
                 layer: layer,
                 dataset: dataset,
                 mouse: d.mouse,
-                datum: d.data
+                datum: d.data.properties
             });
 
             if (!this.clickInfoWidget) {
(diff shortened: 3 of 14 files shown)


aashish24:
@jbeezley this is awesome stuff :fireworks: :fireworks: :fireworks:  Will have a look at it

aashish24:
> Performance: the performance of the polygon rendering is pretty terrible. There are number of offenders, the worst seems to be triangulating the polygons, which takes about 30 seconds.

Do you know if the library we are using is the culprit here or our code?

aashish24:
> Performance: the performance of the polygon rendering is pretty terrible. There are number of offenders, the worst seems to be triangulating the polygons, which takes about 30 seconds.

We could possibly do this on the backend but will require VTK or third party library.

jbeezley:
It spends almost 100% of the time in [this function](https://github.com/jahting/pnltri.js/blob/72f162d39da3a84c78ff16181940da79f3f50ba9/src/EarClipTriangulator.js#L41).  I'm guessing the `O(n^2)` is what's at issue.

aashish24:
> It spends almost 100% of the time in this function. I'm guessing the O(n^2) is what's at issue.

thanks, that confirmed my suspicion. I will file an issue for it in GeoJS (if you haven't doen it) but it could be hard problem to solve in JS world. As discussed one viable option would be to use the converted dataset (assuming its not too big) for visualization.

jbeezley:
After exploring the performance issues a bit more, it appears as if the triangulation is mostly a problem only when profiling it.  When you load the page without the debugger open it only takes a second or two to render the polygons.  Here is a minimal example of what I'm talking about: http://jsbin.com/xoriqub/2.

The triangulation is definitely a good candidate for processing in a web worker, or at least to perform in batches asynchronously.  For the moment, I think we can leave it as is.

aashish24:
> After exploring the performance issues a bit more, it appears as if the triangulation is mostly a problem only when profiling it. When you load the page without the debugger open it only takes a second or two to render the polygons. Here is a minimal example of what I'm talking about: http://jsbin.com/xoriqub/2.

Cool.. yes 30 seconds surprised me since I know that when I was testing it, it seems quite fast (and it claims to be). 

> The triangulation is definitely a good candidate for processing in a web worker, or at least to perform in batches asynchronously.

+1. Do you want to create an issue for it?

Also, can we apply some transparency by default may be have it 0.75 opacity by default?

jbeezley:
Color schemes from colorbrewer have now been added.
screen shot 2016-01-14 at 2 42 20 pm

aashish24:
> Color schemes from colorbrewer have now been added.

and they are so pretty :eyes:

jbeezley:
@mgrauer @aashish24 I'll call this ready for review now.  I believe everything in the [MMWR milestone](https://github.com/Kitware/minerva/milestones/Fetch%20MMWR%20data%20and%20visualize%20summary%20in%20Minerva) is now done.

jbeezley:
Final screenshot showing the info widget that pops up when you click on a state:
screen shot 2016-01-15 at 3 20 14 pm

web_external/js/views/body/MapPanel.js
```diff
@@ -111,6 +198,12 @@ minerva.views.MapPanel = minerva.views.Panel.extend({
                 }
                 this.uiLayer = this.map.createLayer('ui');
                 this.map.draw();
+            } else if (dataset.getMinervaMetadata().source === 'mmwr_data_import') {
+                // hacktastic special handling of MMWR data
```
mgrauer: this is well in line with our current design

Gruntfile.js
```diff
@@ -76,7 +76,10 @@ module.exports = function (grunt) {
         stylus: {
             minerva: {
                 files: [{
-                    src: ['<%= plugin.minerva.external %>/stylesheets/main.styl'],
+                    src: [
+                        '<%= plugin.minerva.external %>/stylesheets/main.styl',
+                        '<%= plugin.minerva.root %>/node_modules/colorbrewer/colorbrewer.css'
```
mgrauer: This would be better in some target that runs on grunt init, similar with the js below.

mgrauer:
@jbeezley 

I'm not sure if you are wanting to merge this branch or just use it as a demo, or maybe just get it in and clean it up later.  I'll wait on further QA until you let me know.

I tested it, it works and looks very nice.  

Will this play well with GeoJs 0.6.0?  I'm assuming so.  I plan to update to that very soon.

aashish24:
Great @jbeezley. @mgrauer let me review it first thing in the morning tomorrow.

jbeezley:
@mgrauer  My only concern about this PR is that it continues to bloat the technical debt in this repo.  Every bit of untested, haphazard code like this that we add will make the inevitable refactoring more difficult.  I'll leave the ultimate decision to you and @aashish24.

There shouldn't be any issues with GeoJS 0.6.  I intentionally left out the option of customizing the opacity of the choropleth because upgrading GeoJS will allow us to do so in a much more general manner.

aashish24:
> Every bit of untested, haphazard code like this that we add will make the inevitable refactoring more difficult.

@jbeezley what would it take to do add testing for this new functionality? Are you talking about client side regression testing or in general about testing.

danlamanna:
I think the most "hacktastic" (thanks @jbeezley) portion here (and in Minerva in general) is probably [this beaut](https://github.com/Kitware/minerva/blob/configurable-choropleth/web_external/js/views/body/MapPanel.js#L154). [Though we need more testing on the client side in general](https://github.com/Kitware/minerva/blob/master/plugin_tests/minervaSpec.js#L10), one of the main issues is #157 which is what leads to this "just add another clause to the mapPanel ifelse" business.

jbeezley:
@aashish24  Client side unit testing, ideally.
@danlamanna  Exactly.

There are a couple of reasons I can't just throw in some unit tests onto this PR.  One of which is that we need to update GeoJS so that I can render the map panel without raising an exception due to missing webgl (a related issue is that the polygon feature only exists with gl).  Further without more structure in the code itself, it is difficult to isolate portions or it for the purposes of unit testing.

aashish24:
> @aashish24 Client side unit testing, ideally.

Great..let's add a task related to this in the next sprint of the project. The goal would be to add update / add missing pieces for unit testing.

Gruntfile.js
jbeezley: I'm building the colorbrewer components into the core minerva source bundles.  The biggest reason I did so is because I didn't want add new files that have to be included everywhere (and I guess I'm lazy).  From what I could tell, there is no equivalent to Girder's `ext.min.js` to put external dependencies into a single place.  Should I do something like that here?

mgrauer: No, I'm doing something like that for bringing in datatables.  I'll move these there on that PR.

server/jobs/soda_import_worker.py
```diff
@@ -25,8 +25,8 @@
 
 #: geojson of state boundaries source url
 states_source = (
-    'https://gist.githubusercontent.com/jbeezley/82150e34cee8512815cf/'
-    'raw/32549660c18149698a849b00cac6df8ddd2d1f1a/us-states.json'
+    'https://gist.githubusercontent.com/jbeezley/ec21c4a016a84c5def74/'
```
mgrauer: could you move this into minerva's repo?

mgrauer:
@jbeezley Other than the question about the gist/repo, if you rebase this, I'm fine with it going in.

server/jobs/soda_import_worker.py
jbeezley: If you really want me to, I will.  I was hesitant to do so because it is 2.5 Mb.  I don't like committing files that large to git.  Plus, there will almost certainly be a need for other similar datasets that will definitely be too large to commit.  I'm not sure how we want to handle this problem when it comes to that.

mgrauer: Gotcha, ok, leave it as is.  Thanks for the explanation.

I'll file an issue.

mgrauer:
LGTM.

PR #257, opened 2016-01-14
Dataset uploader

mgrauer:


mgrauer:
Ready for review.  Fixes a few small nags along with adding an uploader to the Dataset panel.  Also cleans up the createDataset endpoint to be better named as
createItemDataset with route minerva_dataset/id/item, since what it does is create datasets out of items.

server/rest/dataset.py
```diff
@@ -299,32 +299,23 @@ def getDataset(self, item, params):
...
-        # 1) get all the files
-        # 2) find their types based on mimetype and extension
-        # one of (geojson, json, csv)
-        # set the original_type and original_file[] meta.minerva fields
-        # with geojson_file if appropriate
-        #
-        # output should be the same the the get returns
-        # which is a blob of minerva dataset metadata
-        #
-        # TODO this switching is based on which file is found first
```
danlamanna: It seems as though this TODO was removed without being cmpleted?

server/rest/dataset.py
```diff
@@ -299,32 +299,23 @@ def getDataset(self, item, params):
...
-        # fairly brittle and should only be called after first upload
-        # perhaps check if this metadata already exists and don't run if so?
-        minerva_metadata = {}
+    def createItemDataset(self, item, params):
```
danlamanna: Can a docstring be added here, or at least restore the removed comment?

web_external/js/views/body/DataPanel.js
```diff
@@ -1,11 +1,72 @@
...
+        var zeroethFileName = null;
+        this.newItemName = null;
+        this.newItemExt = null;
+        if (files && files.length > 0) {
+            zeroethFileName = files[0].name;
+            this.newItemName = zeroethFileName;
+            this.newItemExt = zeroethFileName.substr(zeroethFileName.lastIndexOf('.'), zeroethFileName.length);
+            this.newItemType = files[0].type;
+        }
+    },
```
danlamanna: Can this be written more expressively and compactly like:
```
filesSelected: function (files) {
    if (_.size(files)) {
	   this.newItemName = _.first(files).name;
	   this.newItemExt = _.last(this.newItemName.split('.'));
	   this.newItemType = _.first(files).type;
    }
}
```

danlamanna:
This only works with one file, do we want users to be able to upload more than one dataset at a time or does it really not matter (at least for now)?

I pushed a commit that got rid of the `filesSelected` method, feel free to revert it if you think it's necessary.

Otherwise, this LGTM once the nitpicks I mentioned are resolved and travis is happy.

mgrauer:
I've found some further issues when uploading geojson, it won't display as a dataset.

If I upload, then after it becomes a dataset I refresh, I can display the geojson.  Possibly this is tied in with the GeoJs 0.6.0 upgrade, I'll investigate further later.

Also found an issue displaying contour json.

danlamanna:
@mgrauer I didn't test any integration with the actual map because 1) I assumed as long as the dataset was getting uploaded properly it was fine and 2) I have no example files to run :P

mgrauer:
Good points @danlamanna .  I didn't want to lose that bug report, though it may be appropriate to split out.  I'll be sure to provide you test data if we'll want to fix these issues on this branch.

I'll respond to your QA later today.

mgrauer:
@danlamanna I'm done with this branch, if you can look at it again.

I agree, this only works for one file.  We are likely going to need a more complex wizard for uploads, where users can specify info (e.g., is this geojson or contour json?, what to name the overall item), but we don't have specific use cases yet.

Contour json can be addressed in #272 .

aashish24:
@mgrauer would it be okay if I review this before we merge it?

mgrauer:
@aashish24 

please do.

I forgot to mention where test data lives.  There is geojson and contour json in plugin_tests/data

danlamanna:
This LGTM, @mgrauer the only thing I noticed is if I uploaded points.json and tried to visualize it everything balked - I'm assuming the eyeball is conditional?

aashish24:
@mgrauer nice work! but there are few things I would like to revisit w.r.t to this branch.  Lets talk about them on a hangout if that's okay.

mgrauer:
@danlamanna 

Yes, the eyeball is conditional, and there isn't enough info on (I'm assuming) `two_points.json` to know that it is geojson.  I can add this #272.  If you have specific other guidance for the more complex/widardy upload widget, can you add info there?

@aashish24 

Can you put your comments here?  I'd rather capture things in writing and in GH if possible.

mgrauer:
Well, I've blown out the scope here doing some preparatory* work for #268, which aims to simplify source and dataset by largely combining them into dataset.

These additional changes clean out some old prototype-y code that isn't clearly used anymore, and introduces an API on the dataset model related to GeoJs rendering.  This may not be the ideal location for it, but I think it is an improvement and could be moved later once a better place is figured out.

There's also some protection (finally) for datasets that won't render properly in the map.  If this happens, such as because we guess the wrong type of file, an attention icon is displayed in the layers panel.  In this example I took a `geojson` file and changed the extension to `json`, which at first was attempted to be rendered as contour json.

screen shot 2016-01-22 at 4 08 23 pm

@jbeezley added a cog icon for configuration of choropleth datasets, I piggybacked on this and display the choropleth dialog for those that need it and allow certain other types to change between contour and geojson.  Datasets that are known to have map rendering issues still have attention called to them.

screen shot 2016-01-22 at 4 08 34 pm

The dialog allows you to change their rendering type

screen shot 2016-01-22 at 4 08 47 pm

and then the error attention goes away and the dataset can now be rendered as the `geojson` it actually is.

screen shot 2016-01-22 at 4 13 23 pm



*I got the spelling right on the fourth try.

mgrauer:
Fixes #272.

aashish24:
Yay! @mgrauer is this ready for review?

mgrauer:
Yes, ready for review.

aashish24:
Awesome @mgrauer will do

web_external/js/models/DatasetModel.js
```diff
@@ -66,313 +39,153 @@ minerva.models.DatasetModel = minerva.models.MinervaModel.extend({
...
+                // Guess contour json as a default for a json file.
+                mm.geo_render = {
+                    type: 'contour',
+                    file_id: mm.original_files[0]._id
+                };
+            } else if (mm.dataset_type === 'wms') {
+                mm.geo_render = {
+                    type: 'wms'
+                };
+            }
```
danlamanna: Should we throw a console warning if it's not a valid geo render type here?

web_external/js/models/DatasetModel.js
```diff
@@ -66,313 +39,153 @@ minerva.models.DatasetModel = minerva.models.MinervaModel.extend({
...
+    ////////////////////////////////////////////////////////
+
+    /**
+     * Initialize the GeoJs rendering type along with any secondary data needed for
+     * GeoJs rendering, inferring the rendering type based on
+     * the dataset's metadata, and saving the dataset's metadata if changed.
+     *
+     * @returns {Object} Updated minerva metadata of this dataset.
+     */
+    _initGeoRender: function () {
```
danlamanna: It seems like this is something that should be part of the models initialize method, no? That way the rest of the methods can assume it's loaded already.

web_external/js/models/DatasetModel.js
```diff
@@ -66,313 +39,153 @@ minerva.models.DatasetModel = minerva.models.MinervaModel.extend({
...
-            processCSVFile();
+    /**
+     * Gets the download URL for the file data needed by GeoJs to render
+     * this dataset, if one exists; initializing it if
+     * necessary by inference from the dataset's metadata, and saving
+     * the dataset's metadata if changed.
+     *
+     * @returns {String|null} Download URL for the file data, if one exists.
+     */
+    _getGeoRenderDownloadUrl: function () {
```
danlamanna: This is duplication of what Girder provides in its base model/file model (see girder.Model.downloadUrl).

web_external/js/models/DatasetModel.js
```diff
@@ -66,313 +39,153 @@ minerva.models.DatasetModel = minerva.models.MinervaModel.extend({
...
-            }
-        }, this)).error(_.bind(function (err) {
-            console.error(err);
-            girder.events.trigger('g:alert', {
-                icon: 'cancel',
-                text: 'Could not geocode tweets.',
-                type: 'error',
-                timeout: 4000
+            var url = this._getGeoRenderDownloadUrl();
+            $.ajax({
```
danlamanna: Why `$.ajax` and not `girder.restRequest`? I'm not fully aware of the context in which this is running, but I think this side steps Girder authentication?

web_external/js/views/body/DataPanel.js
```diff
@@ -1,61 +1,101 @@
...
                 });
-                this.tableWidget.render();
-            }, this);
-            dataset.getCSVFile(filesCallback);
+                this.jsonConfigWidget.render();
+            } else {
+                this.jsonConfigWidget.setCurrentDataset(dataset);
+            }
+        } else {
+            return;
```
danlamanna: Can be removed.

web_external/js/views/body/DataPanel.js
```diff
@@ -1,61 +1,101 @@
...
-                dataset: dataset,
-                parentView: this
-            });
-            this.keymapWidget.render();
-        } else {
-            // assuming csv
-            // list the files of this item
-            var filesCallback = _.bind(function () {
-                this.tableWidget = new minerva.views.TableWidget({
+        if (geoRenderType === 'choropleth') {
```
danlamanna: Can there just be a mapping of render types -> widgets to avoid the similar code?

danlamanna:
A couple of things @mgrauer 

1) Why can't we have a GeoDatasetModel? This seems to conflate things between dataset models and geospatial dataset models quite a bit. It looks like it was but then you switched, i.e. `isRenderable` on the `GeojsonDatasetModel` went to being an `isGeoRenderable` method on the `DatasetModel`.
2) `_initGeoRender` and `overrideGeoRenderType` seem like they should/could be one and the same.

web_external/js/models/DatasetModel.js
mgrauer: will do

mgrauer:
Also to add @ fires and documentation on the catalog of dataset types.

web_external/js/models/DatasetModel.js
mgrauer: I can't do it in the initialize method because the metadata won't be there yet, only is there after createDataset (renamed to promoteToDataset) is called.  I'll centralize this so that it is only called from promoteToDataset.

web_external/js/models/DatasetModel.js
mgrauer: I think this is a slightly different situation and can't use the girder.Model.downloadUrl.

That method is for downloading the entire resource, so in the case of an item would download all files.

In this case, I just want to download a single file, and it is a specific file related to geo rendering.

mgrauer: I ended up removing it anyway, you'll see.

web_external/js/models/DatasetModel.js
mgrauer: I'm switching it over to use `girder.restRequest`.  I played with using a FileModel, but I couldn't get it to do what I want.  The FileModel would be good for getting the file document, but less helpful for getting the bytes of the file itself.

You can do

    var fileModel = new girder.models.FileModel({_id: mm.geo_render.file_id});
    var url = fileModel.downloadUrl();

but you can't effectively pass a `url` the restRequest (it fails if you don't pass a path, and if you are going to pass the url and the path, that is confusing and further, what is the point?).

I think that for what I'm doing, using the FileModel only has the advantage of constructing the download URL for me, and this other breakage prevents me from fully exploiting it.

mgrauer:
@danlamanna Ready for review.  I implemented everything you asked, except related to the FileModel, which I commented on above.

>1) Why can't we have a GeoDatasetModel? This seems to conflate things between dataset models and geospatial dataset models quite a bit. It looks like it was but then you switched, i.e. isRenderable on the GeojsonDatasetModel went to being an isGeoRenderable method on the DatasetModel.

Good question, I think this PR doesn't answer this question, but this depends on some time focused to refactor the rendering, as discussed in #157.  

To your specific question, I don't think we should have a GeojsonDatasetModel, because what if we have a dataset that can render both as a time series and on a map, would we have a GeojsonTimeSeriesDM in addition to a GeojsonDM and a TimeSeriesDM?  

I think this PR at least helps corral some of the ugliness related to not yet solving #157 to fewer locations.

danlamanna:
@mgrauer Looks like some legitimate test failures. Once those are fixed, this LGTM.

mgrauer:
:snowflake:

PR #259, opened 2016-01-16
Upgrade Girder, Minerva, Romanesco common requirements

brianhelba:
cffi changes at:
https://cffi.readthedocs.org/en/latest/whatsnew.html#v1-5-0

GeoJSON changes at:
https://github.com/frewsxcv/python-geojson/blob/master/CHANGELOG.rst

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,4 +1,4 @@
+geopandas==0.1.1
 netCDF4==1.1.8
 Sphinx==1.2.2
 sphinx_rtd_theme==0.1.6
-geopandas==0.1.1
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,4 +1,4 @@
 geopandas==0.1.1
 netCDF4==1.1.8
-Sphinx==1.2.2
-sphinx_rtd_theme==0.1.6
+Sphinx==1.3.5
+sphinx_rtd_theme==0.1.9
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,6 @@
-cffi==1.3.0
+cffi==1.5.0
 cryptography==1.1.0
-geojson==1.1.0
+geojson==1.3.2
 geopy==1.10.0
 girder-client==1.1.0
 ijson==2.2.0
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,10 +1,10 @@
-girder-client==1.1.0
-jsonpath-rw==1.4.0
-geojson==1.1.0
-ijson==2.2.0
-geopy==1.10.0
-owslib==0.9.1
 cffi==1.3.0
 cryptography==1.1.0
+geojson==1.1.0
+geopy==1.10.0
+girder-client==1.1.0
+ijson==2.2.0
+jsonpath-rw==1.4.0
 numpy==1.10.1
+owslib==0.9.1
 python-dateutil===2.4.2


brianhelba:
This should be merged along with: https://github.com/girder/girder/pull/1177 and https://github.com/Kitware/romanesco/pull/203 to bring all three projects in sync and up to date (with respect to common requirements only).

mgrauer:
These changes seem reasonable to me, but I have a few concerns.

Minerva is pegged to specific versions of Girder and Romanesco, so if you are upgrading the 3 projects together to have common requirement versions, that won't necessarily work for Minerva, as it isn't run against the master version of the other two and this could cause compatibility problems as we upgrade the Minerva versions of these dependencies but still run against older non-upgraded versions of the Python requirements in Girder and Romanesco.

@jbeezley knows a lot more about that than me, so I'll be happy to follow his advice there.

Also, we need to figure out a policy for testing ansible/vagrant provisioning, which @kotfic very recently got working.  For something like changing the requirements file, we should have the QA person provision a Minerva VM setting the [minerva_version](https://github.com/Kitware/minerva/blob/9181054f04a2c839716b5c26c60fe23dcddba00d/ansible/group_vars/girder.yml#L11) to the proposed branch name--I will do this now as QA on this branch and report back.

If other Minerva devs agree with this, we should add the policy to the developer docs.

mgrauer:
I tried installing this with ansible/vagrant and ran into trouble.

We might have to wait for the next release of girder and romanesco to bring this in.

>TASK: [Install minerva and romanesco as girder plugins] ***********************
failed: [minervgrant] => {"changed": true, "cmd": ". /opt/.girderenv/bin/activate && girder-install plugin -s -f /opt/minerva /opt/romanesco", "delta": "0:00:00.063277", "end": "2016-01-17 00:07:01.885452", "rc": 1, "start": "2016-01-17 00:07:01.822175", "warnings": []}
stderr: Traceback (most recent call last):
  File "/opt/.girderenv/bin/girder-install", line 5, in <module>
    from pkg_resources import load_entry_point
  File "/opt/.girderenv/local/lib/python2.7/site-packages/pkg_resources.py", line 2720, in <module>
    parse_requirements(__requires__), Environment()
  File "/opt/.girderenv/local/lib/python2.7/site-packages/pkg_resources.py", line 588, in resolve
    raise DistributionNotFound(req)
pkg_resources.DistributionNotFound: cffi==1.3.0

brianhelba:
@mgrauer This should not be a problem.

First, note that prior to this PR, the dependencies between Girder and Minvera were already out of sync, so this shouldn't make anything worse.

The reason why it was working before is because thanks to @jbeezley , [Girder's setup.py](https://github.com/girder/girder/blob/master/setup.py#L60) is very permissive about which versions it accepts. With 2 exceptions, as long as the dependant packages are simply present, Girder's dependencies from ``pip install girder`` are satisfied.

The only potential trouble would come if you tried to ``pip install -r ./girder/requirements.txt && pip install -r ./minerva/requirements.txt``. Prior to this PR, that should fail (or rather, technically succeed, but the second command would overwrite the packages from the first one).

So, while I hope @jbeezley agrees with me, I believe this PR will actually improve the current situation. Minerva does not have a ``setup.py``, but if one is ever added, it should use permissively-versioned abstract dependencies, like Girder.

I agree that ``vagrant up`` is a decent QA step to run on any PR, so please let me know how it goes.

brianhelba:
@mgrauer Ah, I see the problem with your new Vagrant+Ansible test. The latest release of Girder still has the [old strictly-versioned setup.py](https://github.com/girder/girder/blob/v1.4.1/setup.py#L63).

So, we should cut a new release of Girder before this PR goes in. Interestingly, I think we can actually safely put https://github.com/girder/girder/issues/1177 into that new Girder release, since as I mentioned, the ``requirements.txt`` dependencies were already out of sync on both of the latest master branches.

brianhelba:
@zachmullen when's the next Girder release likely to come?

zachmullen:
Whenever we want, I've been meaning to cut one for a while now.

brianhelba:
Finally, I'm surprised that breakages are not already occurring due to Romanesco. Romanesco has a [requirement for ``girder-client=1.0.1``](https://github.com/Kitware/romanesco/blob/v0.2.0/romanesco/plugins/girder_io/requirements.txt) in both the latest release and master, while and [Minerva requires ``girder-client=1.1.0``](https://github.com/Kitware/minerva/blob/master/requirements.txt#L1). The real problem is not this inconsistency, but that Romanesco has a ``setup.py`` that turns its ``requirements.txt`` directly into the package dependencies.

So, Romanesco should also be fixed to use more abstract unversioned dependences in its ``setup.py``, as Girder already does.

This is, of course, outside the scope of this PR though, as the packages updated within this PR are not shared with Romanesco.

mgrauer:
@brianhelba I'm trying to push this through, as you can see with my additional ansible commits here.  They may not end up going in with this branch in Minerva, but I want to give the go ahead to girder to merge https://github.com/girder/girder/issues/1177 .  I'll let you know any further trouble I run into.

I agree, from the Minerva perspective, there is no reason not to merge that Girder PR, especially as we are pegged to an earlier version of Girder.

Your Romanesco requirements changes required me to add two system packages for Pillow, that should probably be documented somewhere.

@zachmullen any interest in a Romanesco release soon?  We found a breaking change in the last couple PRs, and so we've pegged to https://github.com/Kitware/romanesco/commit/cfdc285fd2d94ac1b8793b33e5ef3b49e59ccbb3 for now.  I'll get you a bug report or question on those breakages soon.

zachmullen:
What's the breaking romanesco change? We should try and fix that...

brianhelba:
@mgrauer Please make a comment over in https://github.com/Kitware/romanesco/pull/203 , as to what extra packages you needed for Pillow ``2.4.0`` -> ``3.1.0``.

mgrauer:
@brianhelba With these additional commits on ansible, which add in the system packages and put all three repos on your well named branch, this fully provisions Minerva.  We can't merge it here, but I give you a LGTM for Girder on Minerva's behalf.  I'll do it there too.

zachmullen:
Let's go ahead and merge this then.

mgrauer:
@zachmullen I'm getting `Uncaught TypeError: Cannot read property 'registerStatus' of undefined` in the Romanesco plugins.min.js.

Though the `girder.jobs_JobStatus.registerStatus` object is defined in the JS context.

```
>girder.jobs_JobStatus.registerStatus
(a){_.each(a,function(a,b){this[b]=a.value,this._map[a.value]={text:a.text,icon:a.icon}},this)}
```

zachmullen:
Is the jobs/plugin.min.js loaded before romanesco/plugin.min.js?

mgrauer:
No, it is loading after Romanesco, so that explains this breakage.

But on an earlier version of Girder (we peg to 1.4.1) and that SHA I mentioned for Romanesco, these resources are loading in the correct order (by dependency).

In Minerva we [loop through the plugins in mako](https://github.com/Kitware/minerva/blob/master/server/minerva.mako#L50) but don't define the order, and haven't changed that for a little while.

Maybe the order in Girder's [webroot.py](https://github.com/girder/girder/blob/1bee162bdbe94b800426b784eba87106e5378dd1/girder/utility/webroot.py#L97) or where it gets that list from changed recently, sometime since 1.4.1?

zachmullen:
I believe in girder we load them in the order returned by our toposort, which should be standard practice.

brianhelba:
@mgrauer @zachmullen Just to be clear, is *this* PR safe to merge now?

mgrauer:
No, we can't merge this one yet because it won't install against the version of Girder we are pegged to.

We'll wait for the next Girder release and any work to update Minerva accordingly, along with tracking down this issue with our use of Romanesco.

@brianhelba is this holding you up somehow?

brianhelba:
@mgrauer It's totally fine to wait. I was unclear on the meaning of some prior comments in this thread about issues being fixed.

mgrauer:
@zachmullen I dug in a bit, and there are two separate issues.

1) Load order for client side code of Girder plugins

This is not the reason for the breakage I mentioned, but also not working as you might think.

You can see in [server.py](https://github.com/girder/girder/blob/a8405815c507e3fb13eaf39b8c897500bd1ac260/girder/utility/server.py#L103-L123) that plugins are loaded out of the DB (and they are not saved in toposorted order there).  The list of plugins is passed to `root.updateHtmlVars`, but the plugins have not been sorted in dependency order until further down, in `plugin_utilities.loadPlugins`.  Girder's [mako template](https://github.com/girder/girder/blob/fda9b8eb847e311f1df852f7d97a05c7b43b3a60/girder/utility/webroot.mako#L25-L27) loads the client resources according to this order passed in the html vars, as does Minerva's mako template.  So plugins are loaded in dependency order on the server but not in the client.

2) Minerva/Romanesco/Girder versioning

This is the reason for the breakage.  When Minerva tries to use the latest Romanesco, that depends on a [commit](https://github.com/girder/girder/commit/b97fbbb3e077adce582a97d4dc90efae8809133d) in a very recent Girder, whereas Minerva is pegged to Girder 1.4.1.

I know we've talked about a plan to version plugins separately from Girder, and also to split off the Girder plugin part of Romanesco and bring it into Girder itself.  Once this is done, Minerva can peg a version of Girder and a version of the Romanesco/Girder-worker plugin.  For now we'll peg to Kitware/romanesco@cfdc285.

mgrauer:
I'm correcting this to say that the error is due to both issues.  Even when I set Girder to master and Romanesco to master, I still have the breakage due to the incorrect ordering of client side js resources.

If I reorder them to dependency order, I don't get a breakage.

@zachmullen when you run current master of girder + romanesco plugin together do you not see this client side breakage?  Am I missing something obvious?

zachmullen:
I don't experience that breakage, but that could be just luck if the order is undefined. I will dig into that issue.

jbeezley:
In regards to dependency versions, the `requirements.txt` files between our related projects should not depend on each other at all.  It is a testing environment only and provides explicit versions so that the test outcomes are repeatable.  What we *should* be doing is what (for the most part) Girder does now: a maximally permissive `setup.py` and a `requirements.txt` file that contains the output of `pip freeze` after all dependencies have been installed.

Girder still does use the `requirements.txt` files in plugins for the purposes of the `girder-install` command.  This is an outstanding issue that needs to be addressed.  Each plugin needs to have its own `setup.py`, and if it has standalone testing then the requirements file should contain a full `pip freeze` that is independent of Girder's and any other plugin.  I didn't do this with my PR because it is an issue of backward compatibility, plus it has more complicated implications in terms of how and where the plugins are installed.

zachmullen:
@mgrauer I found the bug related to ordering. Looking at [this code](https://github.com/girder/girder/blob/master/girder/utility/server.py#L103-L112), we simply set the order in the HTML to be however it's stored inside the setting, i.e. we are assuming it's saved to the database in toposorted order. However, [when saving the plugins enabled setting](https://github.com/girder/girder/blob/master/girder/models/setting.py#L56-L86) we don't properly toposort them. I'm going to add a fix in girder that changes that validation method to toposort the list and save it in a loadable order if that sounds good to you.

mgrauer:
That sounds good @zachmullen .  

Will you include a change to sort the settings in existing instances?  Or we can just let them sort themselves out the next time plugin settings change/save, i.e. say turn a plugin on and off.

zachmullen:
@mgrauer The change I made actually will correctly load the plugins in order even if they weren't stored in order, so no migration should be needed.

zachmullen:
https://github.com/girder/girder/pull/1186

brianhelba:
Pushed more upgrades, corresponding to https://github.com/girder/girder/issues/1209 .

Once Girder and Romanesco's upgrades are completed, the "Provision Girder, Romanesco and Minerva on upgrade-common-deps" commit in this PR should be updated to point to the new commit SHAs for the other repos.

aashish24:
@brianhelba could you please update status of this PR.

jbeezley:
I realized last night that #302 and #311 walked all over this PR.  The changes to `main.yml` may still be needed.  The changes to `site.yml` is no longer desired after #302. The changes to the requirements files is part of what I was trying to avoid in #311.

This does leave an unresolved question though.  How do we handle dependencies in ansible deployments?  A [single monolithic freeze file](https://github.com/Kitware/minerva/blob/master/cmake/travis-pip-freeze.txt) is fine for a testing deployment, but if our goal is to provide pluggable components we get back into dependency hell.

The only way forward that I can think of for the moment is to provide a list of "local only" requirements without strict version restrictions (as done in girder's [`setup.py`](https://github.com/Kitware/girder/blob/master/setup.py#L63-L74)).  Ordinarily, a deployment script would just run `setup.py install`, but girder plugins can't be installed that way.  We could use the requirements files to provide the relaxed versions, but that goes against standard python conventions.

At this point, I'm inclined to recommend removing the version numbers from the requirements files here.  We are working against the python ecosystem with our packaging and deployment anyway, so ignoring tradition on this isn't too awful.

Also, sorry about the tea party.

mgrauer:
@jbeezley 

Can you explain further on 

> Ordinarily, a deployment script would just run setup.py install, but girder plugins can't be installed that way.

?


I made the changes in `site.yml` and `main.yml` long ago to make this branch work, and they do not need to be merged for the same reasons that the Library at Alexandria was burned.  The changes in `main.yml` are redundant (already in master), and those in `site.yml` are heretical (go against what #302 brought in).

aashish24:
> I realized last night that #302 and #311 walked all over this PR.

This is one of thing I mentioned. Long standing PR's is a problem. 



> The changes to main.yml may still be needed. The changes to site.yml is no longer desired after #302. The changes to the requirements files is part of what I was trying to avoid in #311.

@jbeezley can you talk to @brianhelba and may be you update this PR? 


> This does leave an unresolved question though. How do we handle dependencies in ansible deployments? A single monolithic freeze file is fine for a testing deployment, but if our goal is to provide pluggable components we get back into dependency hell.

Let's discuss this in our next Minerva meeting.This is a long standing issue.

jbeezley:
It's fundamentally a girder issue.  All that we can do in minerva is work around it.  We addressed part of it in https://github.com/girder/girder/pull/1138, but that PR only went as far as I could take it while maintaining any sense of backward compatibility.  I have some ideas floating in mind about how to make these deployment issues easier.  I'll try to make some concrete recommendations to the girder developers list when I get a chance.

zachmullen:
> that PR only went as far as I could take it while maintaining any sense of backward compatibility

With 2.0 on the horizon, we have a chance to break compatibility if needed.

aashish24:
> It's fundamentally a girder issue. All that we can do in minerva is work around it. We addressed part of it in girder/girder#1138, but that PR only went as far as I could take it while maintaining any sense of backward compatibility. I have some ideas floating in mind about how to make these deployment issues easier. I'll try to make some concrete recommendations to the girder developers list when I get a chance.

thanks @jbeezley for a detailed explanation. Yes, that would be great.

aashish24:
@brianhelba I am closing this PR, feel free to open it when you have time.

brianhelba:
I'm not clear on the resolution here. This PR was created to resolve the inconstancies in ``requirements.txt`` and ``requirements-dev.txt`` between Girder, [the package formerly known as] Romanesco, and Minerva. @mgrauer asked to hold on merging it until Minerva could be pegged to the correct version of Girder.

Now this is being closed, but versions in Minerva's ``requirements.txt`` and ``requirements-dev.txt`` for the packages ``cffi``, ``geojson``, ``Sphinx``, and ``sphinx_rtd_theme`` still do not match Girder upstream.

Forgive me, because I haven't had a chance to keep up to date with all of @jbeezley 's changes to dependencies in Minerva (though making ``setup.py`` use abstract requirements is definitely a good thing). Regardless, if Minerva is going to include  ``requirements.txt`` and ``requirements-dev.txt`` files, they should be kept in sync with Girder, correct?

If so, I'm fine with waiting to merge this until other changes can go in, but I don't think it should just be closed and abandoned.

brianhelba:
I've rebased on the latest master. @mgrauer @jbeezley is this safe to merge now?

jbeezley:
The tea party is over,  LGTM.

mgrauer:
I'll check with ansible.

mgrauer:
I noticed an issue with vagrant/ansible while testing here.  Mongo's config looks like 

    dbpath=/var/lib/mongodb
    logpath=/var/log/mongodb.log
    logappend=true
    bind_ip=10.0.2.15
    port=27017
    journal=true

I'm used to being able to attach a mongo source into minerva using `mongodb://127.0.0.1:27017/my_db`, but instead (because of the `bind_ip`) I needed to do `mongodb://10.0.2.15:27017/my_db`.  This may not be the worst thing.

ping @kotfic  for thoughts or shrugs.

mgrauer:
LGTM.

PR #260, opened 2016-01-17
Ansible debuggability

mgrauer:
@kotfic  PTAL.

While I was playing with this ansible stuff, at first I tried to set the `minerva_version` in ansible/group_vars/girder.yml, but this is not used by site.yml.  Are the three versions in group_vars/girder.yml used anywhere?  If not, can we remove them, since they can be confusing?

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -9,7 +9,7 @@ Vagrant.configure(2) do |config|
 
   config.vm.network "forwarded_port", guest: 8080, host: host_port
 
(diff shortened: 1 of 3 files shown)


kotfic:
@mgrauer LGTM

```minerva_version``` will be picked up from ansible/group_vars/girder.yml if the minerva_version in the playbook role is removed.  Variables declared on roles take precedence over variables contained in group_vars/  files,  they are play-wide, group-wide defaults.

PR #264, opened 2016-01-18
Allow Minerva to be installed to an aribrary directory

jbeezley:
This corrects the issue found by @kotfic when minerva was installed into a directory that was not named `minerva`.  With this change, the directory name should be arbitrary.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -51,22 +51,20 @@ module.exports = function (grunt) {
     }
 
     grunt.config.merge({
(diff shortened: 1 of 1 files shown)


jbeezley:
On second thought, I'm going to close this.  It broke the current build for some reason (I'm not sure why).  There are a number of other issues I found while debugging that make this issue more complicated that I originally thought.  For now, let's just require minerva exist as `plugins/minerva`.

PR #265, opened 2016-01-18
Update geojs to v0.6.0.

### LINKED ISSUE #239: Prevent map scrolling out past max
Is there a way to know the max scroll value, and prevent the map from going past that?

So we don't end up with views like

screen shot 2015-12-21 at 10 55 17 am

mgrauer:
Fixes #239.

I want to fix the WMS api as part of this, but checking on travis first.

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
   "repository": "https://github.com/Kitware/minerva.git",
   "dependencies": {
     "JSONPath": "0.10.0",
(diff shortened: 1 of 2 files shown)


mgrauer:
@jbeezley  PTAL.

jbeezley:
@mgrauer I made a small change to improve the appearance of WMS layers.  With that it LGTM.

mgrauer:
Thanks, LGTM.

PR #274, opened 2016-01-21
Fix incompatibility with GeoJS 0.6 in WMS layers

### LINKED ISSUE #266: Error when clicking on WMS visualization
I think this was introduced by the MMWR info click work recently, and I noticed it when bringing in GeoJs 0.6.0 (so be sure to have #265 merged in to test).

If you have any WMS layer visualized, if you click on the map, you'll get the following error in the javascript console.

> WMSFeatureInfoWidget.js:81 Uncaught TypeError: Cannot read property 'll2m' of undefined

jbeezley:
Fixes #266, but the feature still doesn't work for me.  There are no errors raised in the console, instead the WMS server responds with a message like this:
```xml
<?xml version="1.0" encoding="UTF-8" standalone="no"?>
<!DOCTYPE ServiceExceptionReport SYSTEM "http://geodata.epidemico.com/geoserver/schemas/wms/1.1.1/WMS_exception_1_1_1.dtd">
<ServiceExceptionReport version="1.1.1" > 
<ServiceException code="LayerNotDefined" locator="layers">
      Could not find layer geonode:CHN_ppp_v2c_2010_UNadj_tiled
</ServiceException>
</ServiceExceptionReport>
```

The name is correct in the request, so there could be credentials problem in proxy component.  It is possible there is something incorrect in my test database.

### DIFF
--- a/web_external/js/views/widgets/WMSFeatureInfoWidget.js
+++ b/web_external/js/views/widgets/WMSFeatureInfoWidget.js
@@ -73,17 +73,12 @@ minerva.views.WmsFeatureInfoWidget = minerva.View.extend({
     getUrl: function (layer_idx, coord) {
 
         var pnt = this.map.gcsToDisplay(coord);
-        var mapBounds = this.map.bounds();
+        var mapBounds = this.map.bounds(undefined, 'EPSG:3857');
 
-        var ne = mapBounds.upperRight;
-        var sw = mapBounds.lowerLeft;
-
-        var neMerc = geo.mercator.ll2m(ne.x, ne.y);
-        var swMerc = geo.mercator.ll2m(sw.x, sw.y);
-        if (swMerc.x > neMerc.x) {
-            neMerc.x = 20037508.34 + (20037508.34 - neMerc.x);
+        if (mapBounds.left > mapBounds.right) {
+            mapBounds.right = 20037508.34 + (20037508.34 - mapBounds.right);
         }
-        var bbox = swMerc.x + ',' + swMerc.y + ',' + neMerc.x + ',' + neMerc.y;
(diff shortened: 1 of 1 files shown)


mgrauer:
Thanks for the fix @jbeezley .  This worked for me with 2 opengeo layers and a wwmr choropleth.  Probably it was something with the layer you tested against.

I noticed when I have the choropleth alone, I need to click 2X to get the feature info to come up, after the info panel comes up though I only need one click to get info on a different state.  I'll create an issue for unifying the layer info presentation, we can deal with it there.

screen shot 2016-01-24 at 1 07 12 pm

mgrauer:
LGTM.

jbeezley:
Is the double click to open the info box repeatable?  I noticed GeoJS will often not fire a click event.  This occurs when rendering operations are taking place and is a side-effect of performance problems in GeoJS.  During these blocking operations, mouse events are not captured.  The only way to solve this problem will be to prevent GeoJS from blocking for extended periods of time.

mgrauer:
Repeatable for me in Chrome.  Doesn't matter if I render and immediately click or render, or wait 30 seconds.  Either way I need to double click.  Also I notice that when I render, remove from the map, and rerender, I can no longer bring up the info widget. 

This isn't to say we need to fix this urgently, just wanting to get a complete report.

mgrauer:
I think the first click is actually getting through, because my cherrypy is serving resources after the first click, but the widget doesn't get rendered until the second click (and additional resources are then served).  No idea if which resources get served when is the issue though.

jbeezley:
Probably the widget is created but not shown on the first click.  I'll have a look tomorrow.

PR #280, opened 2016-01-28
Fix MMWR import for new BSVE data scheme

jbeezley:
@mgrauer This should fix the issue we encountered today.


aashish24:
This is quick! Nice @jbeezley

mgrauer:
I'll QA now.  Looks like there was a pep8 failure.  I can fix unless you have already done so @jbeezley ?

jbeezley:
Have at it.  My concentration moved on to other things.

mgrauer:
LGTM.  Thanks for the fix.  Merging.

PR #277, opened 2016-01-25
Revise mako

mgrauer:
I needed to add in datatables for CSV viewing, so I did some work to address #271 .  Further work could be done to unify the css similarly.

This only works with a relatively recent Girder master, so maybe should wait to go in until v1.4.2, or else I should update the Girder branch in Travis to master.

@jbeezley I didn't touch any of the testing stuff so as to not get in your way.  Let me know if this moves things in the direction you want or if you have specific advice.

PR #281, opened 2016-01-28
Peg to Girder SHA 7872249a3 Romanesco SHA da82cd17

mgrauer:


docs/index.rst
```diff
@@ -11,6 +11,8 @@ What is Minerva?
 
 Minerva is an open source platform for geospatial visualization using `Girder`_ as its datastore.
 
+Currently Minerva is pegged to Girder version `7872249a3`.
```
aashish24: So everytime we need to update the docs as well?

aashish24:
Can we not have the version of Girder show up in Minerva (in About dialog or something like that?)

mgrauer:
This PR is to 

1) update the Girder SHA
2) document the Girder SHA
3) open a discussion about how we can peg to Girder

I won't say currently because we don't document it anywhere, but this PR changes the girder SHA in 3 places: travis, our docs, and the ansible playbook for minerva.

What would be ideal is to grab this all from a single location, or fewer locations, and even better if this was officially supported.  I talked to @zachmullen and he is good with adding a `minimum_girder_version` key to the Girder plugin spec.  We would have to implement this in Girder, I think we could do it piggybacking on the work @danlamanna did recently to show which plugins have unmet dependencies, i.e. a message in the Girder plugins page that says "this plugin requires version X of Girder, which is greater than the current version Y" , or some such.

Bringing it back to this PR, ideally we would have it specified in one location and can read that from travis and ansible and possibly readthedocs (it would be easy enough to point people to our plugin.json file), but for now we have to update it in 3 places.

I'll add an issue for an about dialog/hover/display for the Minerva and Girder versions.

aashish24:
Sounds good @mgrauer so is this ready for merge? or you still want to add some more to it? I am trying to see if we can minimize WIP branches :smile:

mgrauer:
@danlamanna brings up good points.

For clarification, @zachmullen has not offered to do this work, but he supports the idea if we implement it.

Beyond the minimum version, we could have a plugin property that is instead "girder_version", and allow ">=, <=, ==, <" to do things like "girder>=1.0,<2.0".  It would be nice to start versioning minerva more actively to make explicit the version of minerva and its relation to the version of girder, though that could potentially complicate matters.

mgrauer:
@aashish24 To me the concern is more about long standing branches hanging out, and I think the bottleneck there is QA and feedback.

This branch needs more work.  I'll definitely update it to drop WIP when ready.

danlamanna:
Curious if @jbeezley has anything to add about plugins pinning Girder to a version since he's been in the versioning trenches a bit.

jbeezley:
I'm still suffering some latent PTSD.  Pinning a to a version is a necessity for CI reliability, but we need to make sure we make it a priority to support the latest version of Girder if not the latest master.  I don't think keeping them in sync should be too much of an issue because most of us actively develop in both places.

Allowing plugins to specify girder versions (particularly in the `<=` direction) worries me.  I think it's generally okay to have a `>=` requirement in the python world, but when you go the other direction you are creating dependency hell for anyone trying you use it.  I think the only sane thing to do is treat a plugin that doesn't work with the most recent version of girder as broken.  If no one gets around to fixing it, then we abandon it.

aashish24:
> This branch needs more work. I'll definitely update it to drop WIP when ready.

Sure, we will try to review sooner and quicker.

mgrauer:
Tested with vagrant/ansible and works.

Ready for review.

mgrauer:
@kotfic This updates Girder SHA for your use.  Can you review? 

and then I can revise #277  accordingly to be based off of this.

danlamanna:
@mgrauer I need this sha of Girder for #289 - https://github.com/girder/girder/commit/e61477c9c50e33a38c39eae0a00c9becca028589

mgrauer:
@danlamanna I updated to your SHA, though this branch name is goofy now.

works on travis, I provisioned with ansible and it works there too.

LGTY?

docs/index.rst
```diff
@@ -11,6 +11,9 @@ What is Minerva?
 
 Minerva is an open source platform for geospatial visualization using `Girder`_ as its datastore.
 
+Currently Minerva is pegged to Girder version `e61477c`.  To update the version of Girder that Minerva
+depends on, update the version here, in `.travis.yml`, and `gider_version` in `ansible/site.yml`.
```
danlamanna: girder*

danlamanna:
Yes LGTM

aashish24:
yay! One less branch

mgrauer:
thanks for the reminder

:snowflake: :snowboarder: :snowman:

PR #286, opened 2016-02-09
Sleep longer in contour_analysis_test

mgrauer:
@cjh1 PTAL, should take you a maximum of 9 seconds.


mgrauer:
I think you would approve, I'll go ahead and merge.

cjh1:
I would have and i do ...

PR #288, opened 2016-02-09
Fix overly general event selector

danlamanna:
This fixes an issue where if the user had an icon-cancel in one of their panels it could also trigger the panel removal.

### DIFF
--- a/web_external/js/views/body/Panel.js
+++ b/web_external/js/views/body/Panel.js
@@ -1,9 +1,9 @@
 minerva.views.Panel = minerva.View.extend({
     /**
-     * The panel view isn't meant to be instantiated on it's own.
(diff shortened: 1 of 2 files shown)


mgrauer:
This is a good change, generally we should target events to classes of our own.

It stood out to me when I tested this that we are using `confirm` to remove a panel, can you change it to `girder.confirm` like [here](https://github.com/Kitware/minerva/blob/c03d0e34ae735fa677178916051668c5501f6ceb/web_external/js/views/body/SessionView.js#L33-L50) to avoid using a browser dialog?

Also, can you change the `it's` to `its` in the comment above the `events` definition?

danlamanna:
@mgrauer

mgrauer:
Thanks, LGTM.

:cloud:

PR #289, opened 2016-02-09
Cumulus front-end plugin

danlamanna:
@kotfic 
The first contrib Minerva plugin, woo!

This PR depends on #277, #288, girder/girder#1230, and the [ansible-refactor](https://github.com/Kitware/cumulus/tree/ansible-refactor) branch of Kitware/cumulus.

Look at 0ed18f3 for the substance of this plugin.

Basically it adds the ability to add/remove AWS profiles within Girder (via Cumulus endpoints), as well as the ability to create, launch, and terminate clusters on AWS using the Ansible playbooks from gobig.

A lot of the logic is around the notion of being able to refresh/leave the page and come back to a sane state of "what clusters are on, launching, terminating, shutdown, etc". This required using the SSE functionality that Girder provides, it enables this plugin to get notifications when the status of a cluster changes, as well as view the live log output of a cluster as it's doing something.

I think I've ironed out the majority of bugs, but the event logic is a bit complex and I'm sure others are looming so review is welcome.

Screenshots:
selection_001
selection_002
selection_003


plugins/cumulus_frontend/web_client/js/views/body/ComputeResourcePanel.js
```diff
@@ -0,0 +1,102 @@
...
+        'click .m-add-computeresource': 'addComputeResourceDialog',
+        'click .m-cluster-details': function (e) {
+            var resource = this.collection.get($(e.currentTarget).attr('m-resource-id'));
+
+            new minerva.views.ComputeResourceDetailWidget({
+                el: $('#g-dialog-container'),
+                model: resource,
+                parentView: this
+            }).render(true);
+        },
```
kotfic: How do you feel about moving these into functions on the view,  that way they can be called programatically if needed?

danlamanna: It looks like what you're suggesting (mostly) tends to be the convention Minerva follows. I'll switch this and the other views I added to use that convention.

kotfic:
Small nitpick,  but can we add a css class or whatever so hovering over an item in the compute resources list switches the mouse to pointer?  For me I am getting the editable text cursor

danlamanna:
@kotfic Yup, done

kotfic:
@danlamanna once this passes travis this LGTM

aashish24:
Nice work @danlamanna @kotfic @cjh1 @mgrauer :+1: :+1:

PR #290, opened 2016-02-10
[WIP] Move datapanel view logic to view

mgrauer:
Ready for review.

I wanted to make the jade a bit more readable (helps by having shorter lines), and also

*  I also moved any logic related to the icon classes, especially any logic dependent on the state of the dataset, into the view.  I'm curious if others think this is a positive change.
* continued jsdoc expansion, this is hard to argue with
* namespacing event classes
* I kept any fontello icon classes in the jade, as this makes it easier to read the jade and see what icon to expect

### DIFF
--- a/web_external/js/views/body/DataPanel.js
+++ b/web_external/js/views/body/DataPanel.js
@@ -164,13 +164,21 @@ minerva.views.DataPanel = minerva.views.Panel.extend({
         minerva.views.Panel.prototype.initialize.apply(this);
     },
 
-    render: function () {
+    /**
+     * Render the DatasetPanel view.
+     */
+     render: function () {
         var datasets = _.filter(this.collection.models, function (dataset) {
             return dataset.metadata();
         });
 
-        // Utility function to control dataset display name.
-        var getDisplayName = _.bind(function (dataset) {
+        /**
+         * Utility function to control dataset display name.
+         *
+         * @param {dataset} dataset the dataset model
+         * @returns {string} display name for the passed in dataset.
+         */
+         var getDisplayName = _.bind(function (dataset) {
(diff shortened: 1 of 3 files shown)


kotfic:
I'm wondering if it makes sense to move the functions like ```getGeoRenderingClasses``` and ```getDatasetName``` into a generic dataset view -  that way the DataPanel doesn't need to know about the specifics of rendering different types of dataset models? Currently if some new dataset model is added that needs custom functionality,  the code for rendering the UI for that functionality has to go into the DataPanel view which mixes concerns. So I guess i'm proposing that the dataPanel template is just responsible for providing a DOM element to append too and rendering is passed off to a generic view that can be extended as needed for custom display functionality.

mgrauer:
@kotfic 

* Add a DatasetView.js/jade/styl for rendering individual dataset rows
* These functions go in that view
* DataPanel hands off the DOM element to each individual DatasetView for each row

Then we can rerender individual dataset rows if need be, rather than the entire Dataset Panel.

Is this what you are saying?

Also, I'd like to rename DataPanel to DatasetPanel at that time.  And change the styling from the fontello icons to our own class selectors.  

Though maybe that would start to get tea-partyish, meaning the scope of the PR is blown out.

kotfic:
@mgrauer that is exactly what i am saying.

I would worry about the rename and styles in a different PR.  That seems like it would be a good 'onboarding'  task for a new UI person?

mgrauer:
OK, I'll split this work out and create some small PRs and an issue or two.

mgrauer:
Closing this as I captured everything I wanted from it in other issues.

Leaving out the name change for now, since we may want to revisit the choice of *Panel as view names.

PR #291, opened 2016-02-10
Transition to ESLint

danlamanna:
There's some issue with the Gruntfile being ignored I have to look into, other than that this is set.


mgrauer:
@danlamanna 

How does updating underscore in Minerva play with the underscore version in Girder?

And can you explain how the new version of _ is pulled in?  I don't see any changes to the Gruntfile, and after blowing away node_modules in `girder` and `minerva` dirs, the `girder` node_modules _ version is 1.5.2.  When I check _.VERSION after rebuilding and serving this branch, my console reports 1.5.2.

mgrauer:
Let me clarify what I did.

I had girder on `e61477c` and minerva on this branch.  Then I deleted `girder/node_modules` and `girder/plugins/minerva/node_modules` and ran `npm install` under `girder`.

danlamanna:
I experienced the same clobbering so I reverted that dependency change - I must have not staged it to commit.

jbeezley:
Updating underscore inside the local `package.json` will only make it available to the grunt scripts doing the build.  To provide a different underscore inside the runtime environment, you will need to override how girder builds its `girder.ext` bundle.

mgrauer:
@danlamanna This looks generally sane.

Can you remove the underscore update and resolve merge commits, then I'll take a final QA pass?  Let me know if you run into merge trouble.

danlamanna:
@mgrauer generally sane is what I was going for. Take a look :)

mgrauer:
LGSTM.

Works on all my layers types, I can pop open an analysis panel, and I can re-order layers.

PR #298, opened 2016-02-13
Add a subcommand to generate BSVE authentication headers

jbeezley:
This was a common problem people had at the TIM.  We were already generating it, this just exposes it through the CLI so that I can point others to it on the developers' forum.

### DIFF
--- a/server/utility/bsve/bsve_utility.py
+++ b/server/utility/bsve/bsve_utility.py
@@ -433,6 +433,10 @@ def main():
         default=None
     )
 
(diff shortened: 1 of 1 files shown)


mgrauer:
Very nice, this will be quite useful and can save people a lot of aggravation.  LGTM.

It output the correct header.


    {
      "harbinger-authentication": "apikey=redacted;timestamp=1455397361801;nonce=66858137;signature=redacted"
    }

PR #299, opened 2016-02-13
Update geojs to 0.7.0

mgrauer:
@jbeezley @essamjoubori  or anyone else, PTAL.

To test this, especially if you haven't had a very recent master of Minerva, you'll need to ensure that

* your Girder is on version e61477c
* you have rebuilt the client (npm install)
* you kill your cherrypy server (python -m girder) and restart it (because we recently updated the mako template, and cherrypy needs to be recycled to serve the new template)

If that doesn't work, you can delete the girder/node_modules folder and the girder/plugins/minerva/node_modules folder then rerun npm install.  Expect to wait on this, and for many npm warnings.

Here's a screenshot from this branch showing 

* wms
* mongo
* geojson
* contour
* choropleth

datasets being rendered with the correct version of GeoJs.

screen shot 2016-02-13 at 3 54 01 pm

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
   "repository": "https://github.com/Kitware/minerva.git",
   "dependencies": {
     "JSONPath": "0.10.0",
-    "geojs": "git+https://github.com/OpenGeoscience/geojs.git#v0.6.0",
+    "geojs": "git+https://github.com/OpenGeoscience/geojs.git#v0.7.0",
     "underscore": "~1.5",
     "jade": "1.3.1",
     "colorbrewer": "1.0.0",


kotfic:
@mgrauer LGTM

aashish24:
@mgrauer I guess we won't be able to see the contour stuff in our local repo? But other than that LGTM :+1:

mgrauer:
@aashish24 Not that contour file I screenshotted, but there is a [contour file in the repo](https://github.com/Kitware/minerva/blob/master/plugin_tests/data/expected_contour.json) (used for testing) that you could use to view in Minerva.

aashish24:
Sure sounds good @mgrauer I won't be able to get it running in next few hours but will try it later. I would suggest let's merge it if CI looks good.

Issue #306, opened 2016-02-16
Add the user login/register controls to the UI


mgrauer:
It would be easy enough to drop in Girder's.  They used to exist but were removed in an overly aggressive purge.

aashish24:
@mgrauer should this be a low priority task?

mgrauer:
I wanted to specify some initial tasks for @katelynnobrien and @essamjoubori related to UI work.

aashish24:
Roger that @mgrauer

mgrauer:
This is another one, in order to be usable we need to have login/register controls.  Currently you have to go through girder to log in.  This needs to be in 0.1.0 to be usable.

PR #310, opened 2016-02-16
[WIP] Add header back to layout

mgrauer:
@katelynnobrien @essamjoubori 

This will get you started for #306.

It adds the header back in, with the user login/logout/register (and user account management) components, along with a globe icon that takes you to the sessions list.  

You can see some problems here, like that the user header obscures the session header (they are stacking over each other) and the globe obscures the name of the session.

Similarly, on the session list page (which is going away anyway), the header obscures the session controls at the top.

screen shot 2016-02-16 at 2 51 01 pm
screen shot 2016-02-16 at 2 51 12 pm


mgrauer:
Closing, this was included in #325.

PR #311, opened 2016-02-16
Submit travis testing to CDash

jbeezley:
http://my.cdash.org/index.php?project=minerva


jbeezley:
I also just pushed a change to install from a "pip freeze" rather than the `InstallPythonRequirements.py` script that we deprecated in girder a while back.  The major benefit here is that numpy will no longer be installed twice.

aashish24:
Very nice @jbeezley

mgrauer:
@jbeezley 

That travis pip freeze file, did it come from Girder+Romanesco+Minerva, all on the correct branches, or how did you make it?

girder: e61477c
romanesco: da82cd17

mgrauer:
Sweet, you included the logo!

screen shot 2016-02-16 at 4 33 17 pm

jbeezley:
I generated the pip freeze by adding the line `pip freeze` after the python requirements were installed on travis in a temporary branch.  I then deleted the reference to romanesco in the resulting file because it had the git sha in it.  That is why I left [this line](https://github.com/Kitware/minerva/pull/311/files#diff-354f30a63fb0907d4ad57269548329e3L93).  It is not ideal, but since romanesco is transitioning to girder_worker anyway, I didn't want to spent too much time worrying about it.

In terms of the exact versions of the packages specified in this file, it isn't terribly important.  The only thing that is important is that the testing environment is completely reproducible.  We can update the environment from time to time (maybe when we update the girder or romanesco sha), but we shouldn't worry about using the exact versions that girder or any other plugin uses because attempting to do so is intractable.

My hope is that we will get to a point that each of the plugins will be provided with their own `setup,py`, and we can do `pip install minerva girder girder_worker ...; pip freeze > travis_pip_freeze.txt` to regenerate the test environment.  Until that happens, we will need to do a little bit more work in curating it manually.

mgrauer:
Thanks for the detailed explanation.

LGTM.


> we shouldn't worry ... because attempting to do so is intractable.

Wise words, though worrying about the intractable is what I do constantly.

aashish24:
> Thanks for the detailed explanation.

Ditto, thanks @jbeezley. @mgrauer We may also want a opaque background logo since the colors are not standing out in a blue background.

PR #314, opened 2016-02-19
fix requirements for older versions of pip

katelynnobrien:
Wouldn't work on old version of pip PTAL @mgrauer


aashish24:
LGTM :+1:

PR #315, opened 2016-02-22
Ensure vagrant mongo listens on all addresses

kotfic:
This makes sure mongo can accept connections from 127.0.0.1 which is a
convention in many people's set ups.

@mgrauer this should address your issue here:
https://github.com/Kitware/minerva/pull/259#issuecomment-185916337


mgrauer:
LGTM, thanks.

PR #319, opened 2016-02-23
Save session as item metadata instead of json file

### LINKED ISSUE #317: Redo client side session as item metadata instead of json
Splitting this out from #225, as it is blocking @jbeezley on #267.

mgrauer:
This changes the Session model to save its Minerva metadata in the Girder Item metadata, just as with the other models.  

Note that to QA this and once this goes in, you'll have to delete all of the Items in your Girder user's minerva/session folder as they are incompatible.

@jbeezley PTAL.

Fixes #317.

### DIFF
--- a/web_external/js/MinervaModel.js
+++ b/web_external/js/MinervaModel.js
@@ -9,6 +9,14 @@ minerva.models.MinervaModel = girder.models.ItemModel.extend({
         }
     },
 
+    /*
+     * Utility method to get this model's current minerva metadata as
+     * it exists on the client, updating this model's minerva metadata to the
+     * minervaMetadata param if passed, but will not save the update.
+     *
+     * @param {minervaMetadata} the object to set as this model's minerva metadata if passed.
+     * @returns {object} current minerva metadata of this model.
+     */
     metadata: function (minervaMetadata) {
         if (minervaMetadata) {
             this.setMinervaMetadata(minervaMetadata);
@@ -17,6 +25,11 @@ minerva.models.MinervaModel = girder.models.ItemModel.extend({
         }
     },
 
+    /*
+     * Gets this model's current minerva metadata as it exists on the client.
+     *
+     * @returns {object} current minerva metadata of this model.
+     */
     getMinervaMetadata: function () {
         // for now assume that keys exists and allow exceptions to happen if they don't
         var meta = this.get('meta');
@@ -28,6 +41,12 @@ minerva.models.MinervaModel = girder.models.ItemModel.extend({
         }
     },
 
+    /*
+     * Sets this model's minerva metadata to the passed object, but does not save
+     * the update.
+     *
+     * @param {minervaMetadata} the object to set as this model's minerva metadata.
+     */
     setMinervaMetadata: function (minervaMetadata) {
         this.set('meta', _.extend(this.get('meta') || {}, {minerva: minervaMetadata}));
         if (minervaMetadata.geojson && minervaMetadata.geojson.data) {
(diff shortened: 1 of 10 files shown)


web_external/js/MinervaModel.js
```diff
@@ -9,6 +9,14 @@ minerva.models.MinervaModel = girder.models.ItemModel.extend({
         }
     },
 
+    /*
```
jbeezley: JSDoc style comments need to start with `/**`.

mgrauer:
That was just plain weird.  Added the missing `*`.

web_external/js/models/SessionModel.js
```diff
@@ -42,99 +36,54 @@ minerva.models.SessionModel = girder.models.ItemModel.extend({
...
-        }, this).on('g:error', function () {
-            minerva.router.navigate('sessions', {trigger: true});
-        }, this);
-        girder.models.ItemModel.prototype.fetch.call(this);
+        map.center = {x: -100, y: 36.5};
+        map.zoom = 4;
+        map.features = [];
+        metadata.map = map;
+        this.on('minerva.metadata.saved', function () {
+            this.trigger('minerva.session.saved', this);
```
jbeezley: Is `minerva.` the new `m:`?

One problem with naming events like this is that jquery interprets them as namespaces which behave much like class selectors.  So this is really triggering a `minerva` event with namespaces `.session` and `.saved`.  It will trigger any of the following event handlers:
* `minerva`
* `.session`
* `.saved`
* `minerva.session`
* etc..

If you want to use namespaced events, then a better choice here would be `saved.minerva.metadata` (or `m:saved.minerva.metadata` keeping with girder tradition).

jbeezley:
This is much better!  Other than my concern about event names, this LGTM.

web_external/js/models/SessionModel.js
mgrauer: >Is `minerva.` the new `m:`?

I was hoping for that, as once we start building many apps or many interacting repos it seemed like there was room for collision.  I'm fine with the `m:`.

Not long ago I had cataloged the Girder events and thought it was unclear from the name where many events were originating.  My overall goal was trying to provide more meaningful event names.

I wasn't aware of the jquery namespacing issues, thanks for explaining, though I don't think I understand yet.  I set up the following event handlers

    session.on('minerva.session.saved', function () { console.log('minerva.session.saved'); }, this);
    session.on('minerva.session', function () { console.log('minerva.session'); }, this);
    session.on('minerva.saved', function () { console.log('minerva.saved'); }, this);
    session.on('session.saved', function () { console.log('session.saved'); }, this);
    session.on('.session.saved', function () { console.log('.session.saved'); }, this);
    session.on('session', function () { console.log('session'); }, this);
    session.on('.session', function () { console.log('.session'); }, this);
    session.on('saved', function () { console.log('saved'); }, this);
    session.on('.saved', function () { console.log('.saved'); }, this);

and when I triggered `'minerva.session.saved'` only the `'minerva.session.saved'` handler reported back.  What am I missing here?

jbeezley: Yeah, I think I had it a little backward in my mind.  Try the following:
* `session.trigger('minerva')`
* `session.trigger('minerva.session')`
* `session.trigger('session)`
* `session.trigger('saved.saved)`
* etc.

Interestingly, `trigger('.saved')` doesn't do what I thought.

There are other complications with `$.off` that are probably more serious.  i.e. `$el.off('.session')` will unbind some of those handlers.  The point is that you shouldn't use `.` in event names unless you specifically want to use namespaces.

jbeezley:
If you want to tackle de-namespacing the events later, then this LGTM now.

mgrauer:
@jbeezley 

I tried different combos of triggers/listeners and only could ever see a one-to-one correspondence.  Clearly I don't understand this, and you have convinced me to stay away from event namespaces entirely.

I thoroughly tested this de-namespacing commit.

I can:

* create/list/rename/delete/view sessions
* display new or existing geojson
* display new or existing contour json
* display new or existing choropleth
* display new or existing mongo
* display new or existing wms

jbeezley:
Does this help?  [http://jsbin.com/ralago/edit?js,console](http://jsbin.com/ralago/edit?js,console)

jbeezley:
LGTM

mgrauer:
That helps, thanks.  Backbone always treats them as a literal string.

PR #316, opened 2016-02-22
Get ansible minerva version from current git branch

mgrauer:
This feature will provision the ansible/vagrant version of minerva using the current git branch, to be set as minerva_version.

E.g.  On my host, I'm on git branch `feature_foo_la_la`, and I provision a new VM.  The minerva on that VM will be on git branch `feature_foo_la_la` as well.  If my minerva git branch on my host is `master`, then the VM minerva will be provisioned on `master` as well.

@kotfic 

I got this working while QA-ing #315.  Fixes #303.

PTAL.

PR #312, opened 2016-02-17
Revenge of csv

mgrauer:
@essamjoubori 

Can you take over on this branch?  It contains much of what you put in earlier.

There may be some style errors, and can you fix the issue where it isn't finding a couple image resources (you'll see in the browser console when you upload/display a csv file), but it should be fairly close, and I can stop bugging you soon asking you about CSV!

Do you want to try work out the issues with the infinite scroll feature of datatables, or just punt for now?

PR #320, opened 2016-02-23
Explain in vagrant docs that minerva_version comes from git

mgrauer:
@kotfic 

I forgot to update the docs as part of #316.

PTAL.


kotfic:
LGTM

mgrauer:
I'll merge this later on, after #319 and #312 go in.  Those are both hard enough to QA without complicating matters.

PR #326, opened 2016-02-29
Improve travis caching

jbeezley:
This updates pip on travis to (re)enable binary wheel caching.  It looks like it brings the pip install step down from 6 minutes to 10 seconds.

I also went ahead and cached the installs of cmake and mongo using the scripts girder uses, and I added similar scripts for spark and scala.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,6 +28,9 @@ addons:
     - libnetcdf-dev
 
 before_install:
+    # update pip for binary wheel caching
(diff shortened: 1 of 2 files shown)


aashish24:
> 6 minutes to 10 seconds

This is awesome! :+1: :+1:

mgrauer:
Great, thanks @jbeezley .

LGTM, merging.

PR #334, opened 2016-03-12
Map layers and adapters

mgrauer:
The purpose of this PR is to solve #157, which though that says datasets should determine how they are added to the map, the spirit of the issue is to make it easier and more structured to define new layer types.  This PR will also decouple a dataset type from a layer type so that there is a multiplexing of dataset types to layer types (at least a structure to build out adapters to do the multiplexing).  

The outcome of this PR should also provide a path to calling some (dataset, layer) adapter server side endpoint, but will not do any server side work itself.


web_external/js/views/map/MapAdapters.js
```diff
@@ -0,0 +1,181 @@
...
+    renderable: function () {
+        this.dataset.once('m:dataset_geo_dataLoaded', function () {
+            this.uponGeoDataLoaded();
+        }, this);
+        this.dataset.loadGeoData();
+    }
+});
+
+/** */
+minerva.models.GeojsonMapLayerModel = minerva.models.ReaderMapLayerModel.extend({});
```
kotfic: I'm a little confused here; isn't it the role of the adapter to obfuscate the underlying data type?

mgrauer:
This is ready for review/discussion.  It provides an implementation for Map Adapters of geojson, contour json, and choropleth layers.  If this seems like a promising direction I can do WMS next.

A few things to note

 * The layer adapter definition has a function `collectUserInput` that takes the dataset as a param, so this would be a place where we could add a dialog tree specific to the (dataset, layer) types.  For now we can use this to grab the user input from the dataset where we store it, as in the choropleth example.
 * I've tried to abstract the idea of a Map Container out from the MapPanel, for now this just looks like a few methods on the MapPanel that the layer adapters call.  This may not end up being useful.
 * The design of the adapters/registry could ideally be simpler, but this was the first way I could get it working with the inclusion of Backbone.Events support.  Simplifications welcome.
 * There is a decoupling of the dataset from the layer type.  For now, we grab the layer type from the dataset, but once we have some UI widgetry to select a type of layer, we should be able to adapt any kind of dataset to a layer (assuming support).  Even though we are passing the dataset to the layer (for reasons of callbacks in the click info widget), we pass the data itself separately into the `initLayer` which means the data doesn't need to come directly from the dataset.
 * Currently the adapter gets the data from the dataset on the client, but this setup should provide a relatively easy path to the adapter getting the data from a server side endpoint and passing it to the layer.

web_external/js/views/map/MapAdapters.js
```diff
@@ -0,0 +1,173 @@
...
+
+minerva.representations.ChoroplethMapLayer = minerva.representations.defineMapLayer('choropleth', function () {
+    this.collectUserInput = function (dataset) {
+        return {
+            colorByValue: dataset.getMinervaMetadata().colorByValue,
+            colorScheme: dataset.getMinervaMetadata().colorScheme
+        };
+    },
+
+    this.initLayer = function (mapContainer, dataset, jsonData, visProperties) {
```
aashish24: why do we need four arguments here? dataset and jsonData seem redundant.

mgrauer: My thinking was that this allows us to decouple the dataset from the data itself.  This API is obviously a work in progress and I'm trying to figure out the right approach.

So when you call initLayer, you pass in the dataset which can be used in a callback, for instance a click info widget that needs to use the dataset to get to some details.  

The adapter takes the dataset along with other properties and produces the jsonData, which is the actual data that should be rendered.  With a different (x,y) center, zoom, or some other properties set, the adapter can get different jsonData for the same dataset.  The dataset isn't responsible for pushing the displayed data to the layer, so it is decoupled.  

For now, we actually get the data from the dataset (in a process mediated by the Adapter Registry), but with this API, it doesn't have to be that way and we could get the data from a the server via an adapter.

web_external/js/views/map/MapAdapters.js
```diff
@@ -0,0 +1,211 @@
...
+    this.setOpacity = function (opacity) {
+        this.geoJsLayer.opacity(opacity);
+    },
+
+    this.render = function (mapContainer) {
+        mapContainer.drawMap();
+    }
+});
+
+minerva.representations.JsonReaderMapLayer = minerva.representations.defineMapLayer('geojson', function () {
```
aashish24: minerva.representations.JsonReaderMapLayer --> 
minerva.rendering.geo.geometryRepresentation

web_external/js/views/map/MapAdapters.js
```diff
@@ -0,0 +1,211 @@
...
+            }, this));
+        } catch (err) {
+            console.error('This layer cannot be rendered to the map');
+            console.error(err);
+            this.trigger('m:map_layer_error', this);
+        }
+    }
+}, minerva.representations.MapLayer);
+
+minerva.representations.ContourJsonReaderMapLayer = minerva.representations.defineMapLayer('contour', function () {
```
aashish24: minerva.rendering.geo.contourRepresentation

web_external/js/views/map/MapAdapters.js
```diff
@@ -0,0 +1,211 @@
...
+    this.readerType = 'contourJsonReader'
+}, minerva.representations.JsonReaderMapLayer);
+
+minerva.representations.ChoroplethMapLayer = minerva.representations.defineMapLayer('choropleth', function () {
```
aashish24: minerva.rendering.geo.chotoroplethRepresentation

web_external/js/views/map/MapAdapters.js
```diff
@@ -0,0 +1,211 @@
...
+                    parentView: mapContainer.getMapView()
+                });
+            }
+        }, this));
+        this.trigger('m:map_layer_renderable', this);
+    }
+}, minerva.representations.MapLayer);
+
+
+minerva.representations.WmsMapLayer = minerva.representations.defineMapLayer('wms', function () {
```
aashish24: minerva.rendering.geo.wmsRepresentation

web_external/js/views/map/MapPanel.js
```diff
@@ -0,0 +1,203 @@
...
+                this.datasetLayerReprs[datasetId] = repr;
+                repr.render(this);
+            }, this).once('m:map_adapter_error', function (dataset, layerType) {
+                dataset.set('geoError', true);
+            }, this).once('m:map_adapter_layerError', function (repr) {
+                if (repr) {
+                    repr.deleteLayer(this);
+                    dataset.set('geoError', true);
+                }
+            }, this).createRepresentation(this, dataset, layerType, mapping);
```
aashish24: createRepresentation --> _createRepresentation

web_external/js/views/map/MapPanel.js
```diff
@@ -0,0 +1,203 @@
...
+        // TODO: HACK MoveToBottom method will set the layer's index to 0 and put it under the base map.
+        // Calling moveUp(1) to place it on top of base map
+        if (dataset.get('order') === 'moveToBottom') {
+            this.datasetLayerReprs[dataset.id].moveUp(baseMapZIndex);
+        }
+        this.map.draw();
+    },
+
+    // MapContainer Interface >>
+    deleteGeoJsLayer: function (geoJsLayer) {
```
aashish24: deleteLayer

web_external/js/views/map/MapPanel.js
```diff
@@ -0,0 +1,203 @@
...
+        this.map.deleteLayer(geoJsLayer);
+    },
+
+    createGeoJsLayer: function (geoJsLayerType, properties) {
```
aashish24: createLayer

web_external/js/views/map/MapPanel.js
```diff
@@ -0,0 +1,203 @@
...
+        // Calling moveUp(1) to place it on top of base map
+        if (dataset.get('order') === 'moveToBottom') {
+            this.datasetLayerReprs[dataset.id].moveUp(baseMapZIndex);
+        }
+        this.map.draw();
+    },
+
+    // MapContainer Interface >>
+    deleteGeoJsLayer: function (geoJsLayer) {
+        this.map.deleteLayer(geoJsLayer);
```
aashish24: geoJsLayer --> geoLayer

web_external/js/views/map/MapPanel.js
```diff
@@ -0,0 +1,203 @@
...
+
+    createGeoJsLayer: function (geoJsLayerType, properties) {
+        return this.map.createLayer(geoJsLayerType, properties || {});
+    },
+
+    getMapView: function () {
+        return this;
+    },
+
+    drawMap: function () {
```
aashish24: drawMap --> draw() ?

web_external/js/views/map/MapPanel.js
```diff
@@ -0,0 +1,203 @@
...
+        this.listenTo(this.collection, 'change:order', function (dataset) {
+            if (this.mapCreated) {
+                this.changeLayerZIndex(dataset);
+            }
+        }, this);
+
+        minerva.views.Panel.prototype.initialize.apply(this);
+    },
+
+    renderMap: function () {
```
aashish24: renderMap + drawMap to one thing?

web_external/js/views/map/MapAdapters.js
```diff
@@ -0,0 +1,211 @@
+minerva.core = minerva.core || {};
+
+(function () {
+    function AdapterRegistry() {
+        this.registry = {};
+        this.register = function (layerType, mapLayerDefinition) {
```
aashish24: layerType -> type
mapLayerDefinition --> definition

web_external/js/views/map/MapAdapters.js
```diff
@@ -0,0 +1,211 @@
...
+            this.registry[layerType] = mapLayerDefinition;
+        };
+        this.createRepresentation = function (mapContainer, dataset, layerType, mapping) {
```
aashish24: mapContainer --> container

web_external/js/views/map/MapAdapters.js
```diff
@@ -0,0 +1,212 @@
...
+                // });
+            }
+        };
+    }
+    minerva.core.AdapterRegistry = _.extend(new AdapterRegistry(), Backbone.Events);
+})();
+
+minerva.rendering = minerva.rendering || {};
+minerva.rendering.geo = minerva.rendering || {};
+minerva.rendering.geo.defineMapLayer = function (layerType, layerDefinition, parentDefinition) {
```
aashish24: defineMapLayer --> createRepresentation ?
layerType --> type
layerDefinition --> functor or something similar
parent Definition -> parent

web_external/js/views/map/MapAdapters.js
```diff
@@ -0,0 +1,212 @@
...
+            this.registry[type] = definition;
+        };
+        this._createRepresentation = function (container, dataset, layerType, mapping) {
+            if (layerType === null || !_.has(this.registry, layerType)) {
+                console.error('This dataset cannot be adapted to a map layer of type [' + layerType + '].');
+                this.trigger('m:map_adapter_error', dataset, layerType);
+                return;
+            } else {
+                var adapter = this.registry[layerType];
+                var layerRepr = _.extend(new adapter(), Backbone.Events);
```
mgrauer: @jbeezley Here's where I brought in Backbone.Events, suggestions welcome for a simpler architecture.

web_external/js/views/map/MapAdapters.js
```diff
@@ -0,0 +1,212 @@
...
+minerva.rendering.geo.defineMapLayer = function (layerType, layerDefinition, parentDefinition) {
+    if (parentDefinition) {
+        layerDefinition.prototype = new parentDefinition();
+    }
+    minerva.core.AdapterRegistry.register(layerType, layerDefinition);
+    return layerDefinition;
+}
+
+minerva.rendering.geo.MapRepresentation = minerva.rendering.geo.defineMapLayer('map', function () {
+    this.deleteLayer = function (container) {
```
aashish24: deleteLayer --> delete

web_external/js/views/map/MapAdapters.js
```diff
@@ -0,0 +1,212 @@
...
+
+    this.render = function (container) {
+        container.renderMap();
+    }
+});
+
+minerva.rendering.geo.GeometryRepresentation = minerva.rendering.geo.defineMapLayer('geojson', function () {
+    this.readerType = 'jsonReader',
+
+    this.initLayer = function (container, dataset, data, visProperties) {
```
aashish24: initLayer --> init

aashish24: We may need exit method as well (which will be called when container gets deleted for example).

web_external/js/views/map/MapAdapters.js
```diff
@@ -0,0 +1,212 @@
+minerva.core = minerva.core || {};
+
```
aashish24: MapAdapters.js --> Adapter.js
views --> should move to core?

mgrauer:
Responded to feedback, ready for review.  @jbeezley this was possibly blocking you, perhaps you could look?

I'm also ok with getting this in sooner rather than later, to tweak it through usage and awkwardness, rather than trying to get it perfect beforehand.

jbeezley:
I haven't followed the discussion about this all that closely, so I don't how deep of a review I can do.  The changes are a clear improvement though.

One thing I would like to see is a description in comments about the API you are creating.  For example, what methods need to be implemented by adapters, what arguments do they take, and what events do they fire?

The `visProperties` objects are where my current work comes in, so they are the focus of my interest.  Do you have any particular plans or ideas about what this object will look like in general?  Do you expect a common interface or will each adapter have a completely independent way of handling styling properties?  Finally, should this mapping be considered mutable, i.e. can you change the style in place without deleting and re-adding a layer?

mgrauer:
@jbeezley 

>One thing I would like to see is a description in comments about the API you are creating. For example, what methods need to be implemented by adapters, what arguments do they take, and what events do they fire?

I was waiting for a bit of stability to do this--I'll do it now.

>The visProperties objects are where my current work comes in, so they are the focus of my interest. Do you have any particular plans or ideas about what this object will look like in general? Do you expect a common interface or will each adapter have a completely independent way of handling styling properties? Finally, should this mapping be considered mutable, i.e. can you change the style in place without deleting and re-adding a layer?

All good questions, with the honest answer being for the most part I don't know, though I'm happy to take direction, especially if you have an actual use case to help guide.  

I expect that each adapter will have an independent way of handling styling properties, though a common API would useful as a way of specifying properties of the visualization (and this could be used in a manual or auto-UI form generation to obtain the vis properties from the user) and getting the specified property mappings within the layer.

I would like it if the mapping is mutable so that you could change the style in place--e.g. you shouldn't have to reparse the entire geojson to change the style properties of a choropleth.  I'm not positive that the current implementation allows this--creating a better API around the `visProperties` object I think will help enable this.  I can also comment in the code to this effect.

jbeezley:
I can probably build on the vis properties in my work.  I was just trying to get an idea where you were going with it.  I think it is probably too much for adapters to share a vis properties "interface", but it would help to have some sort of semantic similarity, maybe through a utility library.

If the vis properties are a mutable in a layer, then we should decide how the layer gets updated.  This will be a much harder problem to solve.  Perhaps the `visProperties` object should trigger change events that the layer listens to?  Should `visProperties` be a full-fledged Backbone model with a rest endpoint and all or just extended from `Backbone.Events`?

mgrauer:
@jbeezley PTAL

I'm punting on making visProperties mutable, and I'm not sure what we'd want there exactly anyway.

I hope that we can get this in after any further QA and improve on it through use.

jbeezley:
Thanks @mgrauer that's exactly what I was looking for.  I should be able to iterate on it from here.  LGTM

mgrauer:
O.

M.

G.

🎊

aashish24:
:fireworks:

PR #335, opened 2016-03-15
Cumulus Provisioning

danlamanna:
@kotfic PTAL


kotfic:
@danlamanna 

I am seeing the following error when trying to create a profile on a 'clean' install running on my local machine -  maybe this is a configuration issue on my part?

```
[2016-03-24 10:28:12,418: INFO/Worker-14] Starting new HTTP connection (1): localhost
[2016-03-24 10:28:12,432: WARNING/Worker-14] {"message": "No matching route for \"PATCH 56f14cc9f4b14909f6151f0b/aws/profiles/56f3f97bf4b1495584c90ffa\"", "type": "rest"}
[2016-03-24 10:28:12,473: ERROR/MainProcess] Task cumulus.aws.ec2.tasks.key.generate_key_pair[d28852a2-991b-4945-a34b-f13fa2530afd] raised unexpected: HTTPError('400 Client Error: Bad Request',)
Traceback (most recent call last):
  File "/home/kotfic/.venvs/NEX/lib/python2.7/site-packages/celery/app/trace.py", line 240, in trace_task
    R = retval = fun(*args, **kwargs)
  File "/home/kotfic/.venvs/NEX/lib/python2.7/site-packages/celery/app/trace.py", line 438, in __protected_call__
    return self.run(*args, **kwargs)
  File "/home/kotfic/kitware/projects/NEX/src/cumulus/cumulus/aws/ec2/tasks/key.py", line 54, in generate_key_pair
    check_status(r)
  File "/home/kotfic/kitware/projects/NEX/src/cumulus/cumulus/common/__init__.py", line 29, in check_status
    request.raise_for_status()
  File "/home/kotfic/.venvs/NEX/lib/python2.7/site-packages/requests/models.py", line 851, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
HTTPError: 400 Client Error: Bad Request
```

More concerning is,  it still creates the profile.  Also - we should disable the 'create profile'  button after clicking until we get a successful return - i clicked twice by accident and got two profiles (both broken because keys were not generated).

kotfic:
@danlamanna we're going to need to add a modal for selecting the play's you want to provision.  This should have a drop-down that is hard-coded for this PR.  long term we're going to want to allow people to search playbooks and add multiple provisioning steps but i think that will probably involve integrating this process with the taskflow plugin.

kotfic:
also we need to figure out why cluster state is showing up as 'undefined' -  server-side code should only be returning strings,  we need to isolate where it is returning numeric codes and identify what is going on server side.

plugins/cumulus_frontend/web_client/js/views/body/ComputeResourcePanel.js
```diff
@@ -21,10 +21,32 @@ minerva.views.ComputeResourcePanel = minerva.views.Panel.extend({
...
+                20:  'launching',
+                30:  'launched',
+                40:  'provisioning',
+                50:  'provisioned',
+                60:  'terminating',
+                70:  'terminated',
+                101: 'stopped',
+                102: 'running'
+            }[status];
+        } else {
```
kotfic: My preference would be to figure out why we're getting status codes instead of strings server-side and open a PR on cumulus

plugins/cumulus_frontend/web_client/js/views/widgets/AddClusterWidget.js
```diff
@@ -21,19 +21,21 @@ minerva.views.AddClusterWidget = minerva.View.extend({
...
-                aws_keyname: this.$('#m-cluster-aws-keyname').val(),
-                instance_type: this.$('#m-cluster-instance-type').val(),
-                instance_count: this.$('#m-cluster-instance-count').val()
+            cluster_config: {
+                master_instance_type: this.$('#m-cluster-instance-type').val(),
+                node_instance_type: this.$('#m-cluster-instance-type').val(),
+                node_instance_count: this.$('#m-cluster-instance-count').val(),
+                master_instance_ami: this.$('#m-cluster-ami').val(),
+                node_instance_ami: this.$('#m-cluster-ami').val(),
+                ansible_ssh_user: this.$('#m-cluster-user').val()
```
kotfic: Would it be possible to include an 'advanced' section in the cluster config popup that allowed adding arbitrary key/value pairs to cluster_config?  Seeing as these get passed through to ansible i could see that being helpful - especially if you need to do something like associate a different python interpreter or something because of your local OS.

danlamanna:
@kotfic Addressed all of the issues you mentioned. This requires latest master from Cumulus, and this has a new `.girder-version` since the merge of Minerva master.

kotfic:
@danlamanna LGTM

PR #340, opened 2016-04-07
Update geojs to version 0.9.0

jbeezley:
jquery-ui was moved out of geojs, so I added as a dependency here.  This uses a different theme so the dialogs look different, otherwise I think this all works the same.


mgrauer:
I had a bit of trouble, but deleting Girder's node_modules and Minerva's node_modules, then npm install worked, I was using Girder v1. 5.0.  Since v1.5.1 is out we should think about upgrading.

LGTM.

Issue #71, opened 2015-08-11
Permissions issue for non-administrative users


kotfic:
Creating a non-administrative user in minerva appears to throw a permissions issue.  As far as I can tell the user does not have access to the session folder that is created for them. 

```
Failed to load resource: the server responded with a status of 403 (Forbidden):8081/static/built/app.min.js:1605
403 Forbidden {"message": "Read access denied for folder 55c9f03cf4b1493c06784ba3 (user 55c9f03cf4b1493c06784b9f).", "type": "access"}defaults.error @ :8081/static/built/app.min.js:1605```

PR #341, opened 2016-04-07
Enable non-admin users

### LINKED ISSUE #165: Enable minerva for non-admin users
Look over #144 for details, but use a different approach.

Currently only admin users can use Minerva, and we are abusing GET requests by creating folders.

Change the approach to dealing with users to be the following

  * when Minerva is enabled as a plugin
    * create a `minerva` collection and a `minerva` group granting read access to that collection
    * create an `analysis` (es?) folder under the `minerva` collection
    * for any existing user
      * create a `minerva` folder under their user
      * create a `dataset` folder under the user `minerva` folder
      * create a `session` folder under the user `minerva` folder
      * create a `source` folder under the user `minerva` folder
      * add the user to the `minerva` group
  * when a new user is added and Minerva is already enabled, listen to the user creation event and follow the above procedure to enable the user for Minerva
  * figure out details related to running romanesco
      * maybe add the user to the list of enabled romanesco users
      * more ideally tie this into the `minerva` group
      * this may involve tying in the `minerva` collection `analysis` folder

mgrauer:
This allows Girder non-admin users to use Minerva.  I don't like that it makes changes on a GET request (really would just be the first time certain folders are needed and then created), but I can live with that.

We still have the sessions page be the first page the user sees, as it will take some thought as to how to remove this page and deal with what to show on the map page when a user isn't logged in.  At least once the user logs in and navigates to an individual session, they can do everything they need to on that session page (including switching sessions and creating new sessions).

Fixes #165, #71.

screen shot 2016-04-07 at 3 33 52 pm

### DIFF
--- a/server/rest/dataset.py
+++ b/server/rest/dataset.py
@@ -266,7 +266,7 @@ class Dataset(Resource):
     @access.public
     @loadmodel(map={'userId': 'user'}, model='user', level=AccessType.READ)
     def getDatasetFolder(self, user, params):
-        folder = findDatasetFolder(self.getCurrentUser(), user)
+        folder = findDatasetFolder(self.getCurrentUser(), user, create=True)
         return {'folder': folder}
     getDatasetFolder.description = (
         Description('Get the minerva dataset folder owned by a user.')
--- a/server/rest/session.py
+++ b/server/rest/session.py
@@ -66,7 +66,7 @@ class Session(Resource):
     @access.public
     @loadmodel(map={'userId': 'user'}, model='user', level=AccessType.READ)
     def getSessionFolder(self, user, params):
-        folder = findSessionFolder(self.getCurrentUser(), user)
+        folder = findSessionFolder(self.getCurrentUser(), user, create=True)
         return {'folder': folder}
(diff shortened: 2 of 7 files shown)


aashish24:
Freedom :smile: How do we test it?

mgrauer:
I need to fix some tests first.  You can test it by registering a new user through minerva and seeing that you don't get errors--you should be able to actually create a new session, goto that session, and do normal things like upload data and render to the map.

aashish24:
Roger that.

mgrauer:
Ready for review.

mgrauer:
@aashish24 This is ready for review.

aashish24:
Thanks @mgrauer I won't get to it today but will try it tomorrow.

kotfic:
@aashish24 This PR is blocking my work so I have reviewed and it LGTM.

I agree @mgrauer that creating folders on GET is a little weird but it is essentially a "lazy load" so i think it's ok.

aashish24:
@kotfic much appreciated, thanks for the review.

PR #346, opened 2016-05-05
Mock wms endpoint for wms auth test

mgrauer:
@kotfic PTAL

I (hopefully) fixed the failing wms test for you.  I mocked the call to the external WMS server, but the test is still valid, the important part is that the credentials were encrypted.

If you think this passes QA, feel free to merge it into master and/or your branch.  I'll be out of touch for the most part and won't be able to merge.


kotfic:
LGTM

Issue #347, opened 2016-05-09
Create setup.py for non-girder related code


kotfic:
Allow minerva to be an install-able package in its own right

PR #348, opened 2016-05-09
Minerva as installable package

kotfic:
This adds a skeleton for minerva as an install-able package.  There are several different components that properly belong outside of the girder environment,  especially in the context of distributed job execution - where workers may need code which cannot otherwise be located by python. 

Currently this does nothing,  but i wanted to get @aashish24 @jbeezley and @mgrauer's review as a separate PR from the analysis PR. If this looks good i will merge and then pull changes into https://github.com/Kitware/minerva/pull/345 before moving forward with remote analysis execution.

ref: https://github.com/Kitware/minerva/issues/347

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,6 @@ src
 /docs/_build/
 .vagrant
 *~
+/*.egg-info/
+/build/
+/dist/
--- a/setup.py
+++ b/setup.py
@@ -1,0 +1,19 @@
+from setuptools import setup
+
+setup(name='minerva.geo',
+      version='0.0.0.dev1',
+      description='Minerva: client/server/services for analysis and visualization',
+      url='https://github.com/kitware/minerva',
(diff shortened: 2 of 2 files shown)


aashish24:
lgtm :+1: (Didn't try to install thus +1 only)

aashish24:
@kotfic ran install as well. I can import minerva, not sure if you want me to verify anything else.

kotfic:
@aashish24 nope,  just wanted to make sure you were happy with the metadata,  if this does get installed on pypi  it will be install-able as minerva.geo.  minerva is taken by another project (https://pypi.python.org/pypi/Minerva/11.7.23)

aashish24:
metadata looks good (anyone else has any suggestions?). I like minerva.geo too.

jbeezley:
LGTM

PR #350, opened 2016-05-10
Fix positions of radio buttons and checkboxes.

manthey:
Bootstrap css expects radio and checkbox input elements to be children of label elements.  If they aren't, the checkbox or radio button is placed 20 pixels too far to the left.  In the dialogs, these are often on or beyond the left edge.


aashish24:
LGTM :+1:

PR #351, opened 2016-05-11
Fix checkboxes in the Wms Layers dialog.

manthey:
In the previous PR, the layout of checkboxes was moved to the Bootstrap motif of including them as children of label elements.  Some code in the Wms layers dialog was then breaking functionality on the checkboxes in that clicking on a checkbox itself would "double" check it, resulting in no change.  This code was designed to prevent clicking labels from activating a disabled checkbox, but is no longer necessary.


aashish24:
thanks @manthey, I will check it out on VM and will report if it works for me (the code change itself) looks fine to me.

aashish24:
LGTM :+1: Verified on a VM

PR #360, opened 2016-05-30
Ui changes

### LINKED ISSUE #357: Some UI suggestions
On the main Minerva App home screen there are 5 tabs on the left hand side of the screen (Analyses, Sources, Datasets, Session Layers and Jobs).

I believe there can be better abstraction between "Session Layers" and "Datasets". I think everything that is about visualizing the datasets such as toggling the layer on and off, changing the opacity, color scheme etc. should be on the "Session Layers" tab. I think in the GIS world natural tendency is to click on the same button while toggling the layers on and off. My proposal is to move the eye button completely to the "Session Layers" and I believe it should work like a checkbox. 

Please see the document:
https://docs.google.com/a/kitware.com/document/d/1wllIZ9iM9VdaU4zfe4odKCdwBwXZHFSh4x4-M5iza1w/edit?usp=sharing

dorukozturk:
This fixes #357.

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,5 +11,8 @@
     "jade": "1.3.1",
     "jquery-ui-bundle": "^1.11.4",
     "underscore": "~1.5"
+  },
+  "devDependencies": {
+    "grunt-contrib-watch": "^1.0.0"
   }
 }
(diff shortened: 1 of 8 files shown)


aashish24:
@dorukozturk I am going to review this branch now.

aashish24:
LGTM :+1: :tada:

PR #362, opened 2016-06-06
Fix vagrant file so vagrant global-status works

kotfic:
Don't assume we're running vagrant commands from same directory as
vagrant file


kotfic:
@mgrauer PTAL thanks!

mgrauer:
LGTM

kotfic:
@mgrauer thanks!

PR #364, opened 2016-06-30
[WIP] Slippy source dataset

aashish24:


aashish24:
@dorukozturk it would be nice if you can try this branch to visualize your raster layers and see if it works - thanks,

PR #366, opened 2016-08-13
Bsve awb sprint1

mgrauer:
Do Not Merge.


aashish24:
@mgrauer should we close this one as well since you are going to create a plugin for this work?

mgrauer:
I'd like to keep it open because it is the easiest way for me to see all the changes that I need to track.  I do not plan to merge it.

PR #368, opened 2016-08-25
Update to geojs 0.10.1

jbeezley:
I briefly tested this.  Someone else should give it a try as well.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -214,7 +214,7 @@ module.exports = function (grunt) {
                     {
                         expand: true,
                         cwd: '<%= plugin.minerva.root %>/node_modules/sinon/pkg',
-                        src: ['sinon-1.17.4.js'],
+                        src: ['sinon-1.17.5.js'],
                         dest: '<%= plugin.minerva.static %>'
                     }
                 ]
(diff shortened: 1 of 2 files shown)


jbeezley:
I also cherry-picked my sinon fix to get the tests to pass.

mgrauer:
@jbeezley When I tried this branch, I can no longer render an uploaded geojson file, but I can do so with current minerva master.

From Adapters.js

>TypeError: Cannot read property '0' of undefined

jbeezley:
For a file with polygons, I was getting a different error.  For some reason, switching to the unminified source for debugging fixed the issue.  There might be something wrong with the minified sources on npm.  See if this works for you.  Otherwise send me the geojson file your using, and I'll do some more exploration.

mgrauer:
Thanks for looking into it @jbeezley, it works now.  Turns out I had an invalid GeoJson file, but one that I had been using for quite a while to test Minerva.  Perhaps GeoJs used to be more tolerant of the invalidity.

LGTM.

jbeezley:
We could probably throw a more useful error, but trying to correct errors in invalid files means slower parsing and more complexity.  The file you were using doesn't load in [geojson.io](http://geojson.io/), and [geojsonlint](http://geojsonlint.com/) can always be used to debug the file.

PR #369, opened 2016-08-29
Fix default tile layer attribution

jbeezley:
The tile url was not being set because it was using the wrong option name, so the OSM tile set was displayed with the mapbox attribution.  This fix will only work for newly created layers.

Also, the mapbox tiles look somewhat nicer, but the response rate is much slower than the OSM server, so I think OSM makes for a better default.


danlamanna:
LGTM

PR #371, opened 2016-08-31
Update girder for upstream deployment fixes

jbeezley:


jbeezley:
Ping.  I think I finally have a working version here.

mgrauer:
I thought we had removed the dependence on Romanesco?

I assume that this failure you are fixing was for the girder-worker core plugin in girder.  Could we also fix it by adding the girder-worker to the

     - export IGNORE_PLUGINS=

line in .travis.yml?

I'm fine with whatever you do, asking to understand it better.

mgrauer:
LGTM

jbeezley:
You're right we could remove girder_worker, but this update was necessary to bring the fontello fix from girder.  https://github.com/girder/girder/pull/1550

PR #372, opened 2016-09-01
Set an arbitrary Girder item as a Minerva GeoJSON dataset

manthey:
Rather than require the item to be uploaded, it can be marked via the POST minerva_dataset_geojson endpoint.

Added a globe icon to the Girder file list to call the endpoint for a specific file within an item.  When a file is used as a Minerva dataset, a globe with an x is shown, clicking which removes the minerva metadata.

If a dataset that is in use in a session is unmarked (by having the minerva metadata tag removed), don't throw javascript exceptions.

The POST endpoint didn't set all of the metadata that was required for sources, etc.  This has been changed, but has not been tested to make sure it doesn't have ill repercussions.

Fixed some CSS that caused Girder to look poorly and require horizontal scrolling.  Also fixed an issue when the main Minerva body would show a vertical scrollbar.

This PR also merged the changes on master.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -162,7 +162,7 @@ module.exports = function (grunt) {
                     {
                         expand: true,
                         cwd: '<%= plugin.minerva.geojs %>',
-                        src: ['geo.min.js'],
+                        src: ['geo.js'],
                         dest: '<%= plugin.minerva.static %>'
                     }
                 ]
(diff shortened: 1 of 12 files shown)


manthey:
@dorukozturk or @aashish24 Did we want to get this reviewed so it can be merged into the sources-refactor branch?

dorukozturk:
LGTM

PR #374, opened 2016-09-07
Point cluster styling

jbeezley:
The styling is pretty basic.  Only constant coloring is possible at the moment.  The clustering distance is a little small, but there is bug in geojs preventing it from being customized.

screen shot 2016-09-07 at 3 45 38 pm
screen shot 2016-09-07 at 3 45 29 pm


aashish24:
@jbeezley thanks for pushing this work. I found some issues: 

1. I had to merge master into it or else I was running into exceptions (not concerned about this issue)

2. When I select clustering option checkbox, I didn't see the clustering options. I had to close and open again. 

3. There is some flickering going on. Not sure if it is a geojs issue or something else. I can show it to you if you want to stop by. 

Thanks,

jbeezley:
Yeah, I just went over it with @dorukozturk, and we noticed something strange about the configuration dialog.  I'll look into it.

I noticed the layers don't stay aligned while navigating.  That is a geojs issue.  I think it is related to https://github.com/OpenGeoscience/geojs/issues/526.  This is using a gl feature layer, but it is doing more processing during navigation events for the clustering computations than most of the gl layers we have tested on before.

aashish24:
> Yeah, I just went over it with @dorukozturk, and we noticed something strange about the configuration dialog. I'll look into it.

Thanks. 

> I noticed the layers don't stay aligned while navigating. That is a geojs issue. I think it is related to OpenGeoscience/geojs#526. This is using a gl feature layer, but it is doing more processing during navigation events for the clustering computations than most of the gl layers we have tested on before.

@manthey can you take a look at it? I remember we talked about it and you were planning at some point to look at it as part of the annotation work.

manthey:
@aashish24 I'll look into the layer alignment in the near future.

aashish24:
@jbeezley just wanted to make sure this is on the list. Let me know when I can test the dialog again. Thanks.

jbeezley:
> When I select clustering option checkbox, I didn't see the clustering options. I had to close and open again.

I think what you are referring to is that the clustering options panel doesn't expand when you click the checkbox.  The option panels expand and contract when you click the headers, not when you enable or disable the option.  This is how it is intended to work, but it might not be the most obvious interface.

aashish24:
> I think what you are referring to is that the clustering options panel doesn't expand when you click the checkbox. The option panels expand and contract when you click the headers, not when you enable or disable the option. This is how it is intended to work, but it might not be the most obvious interface.

Okay, thanks. May be we can follow up on that later and depending on the outcome, do the work in a separate PR.

aashish24:
Works for me!! thanks.

PR #379, opened 2016-11-17
Add the ability to download geojson datasets

jbeezley:


aashish24:
LGTM :+1: and works for me.

PR #380, opened 2016-11-22
Fixed opacity bug

dorukozturk:
This PR aims to fix the inconsistency with the opacity of the layer once it is deleted.


aashish24:
thanks @dorukozturk this would work I think (will try it bit later) but I was wondering if we can force dataset update when it become visible after it gets deleted?

dorukozturk:
@aashish24 I think it is a design decision, do we want to save the opacity state? If yes I think we can update it after deleting and adding it back to the map

aashish24:
I see. I would think that a default of 0.6 - 0.7 might be better but we could take care of defaults in another PR (or a decision of saving the state as well). 

LGTM :+1:

PR #381, opened 2016-11-25
Fix geojson inspect

dorukozturk:
This PR fixes the geojson inspection bug. A new filter is added on client side to stop the querries on geojson layers.


dorukozturk:
First part makes total sense. Shorter the better. I did not quite get the second part though. You think this logic should be moved elsewhere?

dorukozturk:
@danlamanna I changed it according to your comments. PTAL

PR #383, opened 2016-11-28
Collection geojson properties on click

jbeezley:
@dorukozturk This is about half the work required to get geojson property information to display on click.  There are two blocks of code here:

1. Enables the selection api on geojson feature layers with some hackery that is necessary until https://github.com/OpenGeoscience/geojs/pull/649 shows up in a release.
2. Connects to the existing click handler to call the `pointSearch` method on all features created for the dataset.  Because all of data is assumed to be geojson, getting the properties is straightforward.  After the underscore pipeline, I just print out the property array to the console.

What I'm leaving for you is merging these property objects with those returned from the REST API for WMS layers [further down](https://github.com/Kitware/minerva/compare/select-geojson?expand=1#diff-b32150086f8ccbc039b4d05197312c6cR70) in the function.  Let me know if you have any questions.

### DIFF
--- a/web_external/js/views/adapters/Adapters.js
+++ b/web_external/js/views/adapters/Adapters.js
@@ -159,11 +159,11 @@ minerva.rendering.geo.GeometryRepresentation = minerva.rendering.geo.defineMapLa
         var _createFeature = this.geoJsLayer.createFeature;
         this.geoJsLayer.createFeature = function (name, arg) {
             if (!arg) {
-                arg = {}
+                arg = {};
             }
             arg.selectionAPI = true;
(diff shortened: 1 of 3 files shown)


dorukozturk:
@jbeezley Can you review this whenever you have time? I think it is almost ready. When clicked on boundaries it is not inspecting (for geojson only). Other than that everything works as expected.

jbeezley:
My last commit should fix the boundary issue.  Everything else LGTM.

dorukozturk:
Yes, it is fixed now. Thanks !!!

PR #384, opened 2016-12-12
Customize points

dorukozturk:
This PR adds ability to change the marker shape for WMS point layers. Marker options will be only shown to point datasets. 
selection_009

@jbeezley PTAL.

### DIFF
--- a/web_external/js/views/widgets/StyleWmsDatasetWidget.js
+++ b/web_external/js/views/widgets/StyleWmsDatasetWidget.js
@@ -48,7 +48,6 @@ minerva.views.StyleWmsDatasetWidget = minerva.View.extend({
                 this.dataset.fetch();
             }, this));
         },
-
         'submit #m-wms-point-style-form': function (e) {
             e.preventDefault();
             var attribute = this.$('#m-attribute').val();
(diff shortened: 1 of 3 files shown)


web_external/js/views/widgets/StyleWmsDatasetWidget.js
```diff
@@ -48,7 +48,8 @@ minerva.views.StyleWmsDatasetWidget = minerva.View.extend({
                 this.dataset.fetch();
             }, this));
         },
-        'submit #m-wms-style-form': function (e) {
+
+        'submit #m-wms-point-style-form': function (e) {
```
jbeezley: It looks like the submit handler from the old template (`#m-wms-style-form`) will no longer called.

web_external/js/views/widgets/StyleWmsDatasetWidget.js
```diff
@@ -133,8 +136,15 @@ minerva.views.StyleWmsDatasetWidget = minerva.View.extend({
                 max: maximum,
                 ramps: this.ramps
             });
+        } else if (subType === 'point') {
```
jbeezley: Where does this point subtype get set?  In the ingestion, I only see single and multiband subtypes.

dorukozturk: 22783535cf4307505f5bd8b007b5a8e55895d9b4 should fix the submit handler issue

PR #385, opened 2016-12-13
Don't append feature properties if no hits were found

jbeezley:
@dorukozturk This should fix the bug in clicking on geojson layers that we noticed yesterday.


dorukozturk:
LGTM!

PR #388, opened 2017-01-10
Remove BSVE root url code from minerva

jbeezley:
This code is unnecessary as the geoviz plugin computes the base url itself.

### DIFF
--- a/server/rest/feature.py
+++ b/server/rest/feature.py
@@ -59,11 +59,9 @@ class FeatureInfo(Resource):
 
         layerSource = []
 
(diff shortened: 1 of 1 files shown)


jbeezley:
@dorukozturk PTAL

dorukozturk:
@jbeezley So, bsveurl was never used anyways?

jbeezley:
It was, but my minerva-docker PR determines the base url itself.

dorukozturk:
OK, now that makes sense. LGTM

PR #389, opened 2017-01-13
Categorize layers

dorukozturk:
Categorizes layers based on a metadata field that exist in the item. The field has to be in the format of "category:SomeCategory" and the layers then will be categorized based on those fields. If category field does not exist in the metadata, "category:Other" will be used.


web_external/js/views/body/DataPanel.js
```diff
@@ -290,6 +307,13 @@ minerva.views.DataPanel = minerva.views.Panel.extend({
         return (((model.get('meta') || {}).minerva || {}).source || {}).layer_source;
     },
 
+    categorizeLayers: function (source) {
+        var datasetArray = this.sourceDataset[source];
+        this.sourceCategoryDataset[source] = _.groupBy(datasetArray, function (dataset) {
+            return dataset.get('meta').minerva.category;
```
jbeezley: Could you add `|| "Other"` here so it works with old metadata as well?

dorukozturk: Nice catch. I will add it now.

PR #395, opened 2017-03-30
Geojson log scale

dorukozturk:
@jbeezley, @aashish24 PTAL

This is a sample file. Just change the extension to geojson and you should be able to test  the log functionality. There is a new checkbox in the geojson style modal. 

[chickens.txt](https://github.com/Kitware/minerva/files/883407/chickens.txt)

### DIFF
--- a/web_external/js/geojsonUtil.js
+++ b/web_external/js/geojsonUtil.js
@@ -206,9 +206,21 @@ minerva.geojson.style = function style(geojson, visProperties) {
  *
  * @param {string} ramp
  * @param {object} summary
+ * @param {Boolean} logFlag
  * @returns {function}
  */
-minerva.geojson.colorScale = function colorScale(ramp, summary) {
+minerva.geojson.colorScale = function colorScale(ramp, summary, logFlag) {
+    function logarithmicScale(min, max, numBins) {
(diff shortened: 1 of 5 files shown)


aashish24:
thanks @dorukozturk do we have option to use quantize or quantile?

aashish24:
Looks like I think we only have quantize. Can we add quantile as well?

dorukozturk:
@aashish24 I will add the quantile and quantize options in a separate PR, to keep this one small.

web_external/js/geojsonUtil.js
```diff
@@ -212,7 +212,12 @@ minerva.geojson.style = function style(geojson, visProperties) {
  * @returns {array}
  */
 minerva.geojson.logScale = function logScale(min, max, numBins) {
-    var logMin = Math.log(min);
+    // TODO: Figure out what to do when all values are negative
+    var logMin = 0;
+    if (min !== 0) {
```
jbeezley: I would make this `min <= 0`.

jbeezley: Actually, `min > 0`.

dorukozturk: 7df191b applies that.

Issue #398, opened 2017-04-25
Layer mismatching issue


matthewma7:
When a session containing more than one layer loaded, the layers in layer panel would mismatch what's on the map.
See screenshot for detail.

![2017-04-25_12-39-15](https://cloud.githubusercontent.com/assets/3123478/25397041/e401240c-29b4-11e7-8fa7-87cf96071b44.gif)

dorukozturk:
@matthewma7 Should we close this issue?

matthewma7:
Totally, the issue is fixed now. I'll close this now. Thanks! @dorukozturk

PR #399, opened 2017-04-26
Fixes the layer mix matching issue

matthewma7:
This PR aim to fix the issue https://github.com/Kitware/minerva/issues/398

The related events were associated with a singleton object and event is triggered in an async callback, making three different handlers being triggered by one event. However, the code is intended to have three event handler trigger by three different events. 
This PR move events onto individual datasets.


aashish24:
@matthewma7 you can go ahead and merge it

matthewma7:
@aashish24 Cool, thanks!

PR #403, opened 2017-05-11
ENH: Add additional colormap scalings.

danlipsa:
Add clamping for linear scaling
Add quantile scaling
Fix log scaling

### DIFF
--- a/plugin_tests/client/geojsonUtil.js
+++ b/plugin_tests/client/geojsonUtil.js
@@ -326,7 +326,7 @@ describe('geojson', function () {
 
                 expect(scale(0)).toBe(colorbrewer.Reds[9][0]);
             });
-            it('value range', function () {
+            it('linear scale', function () {
                 var scale = minerva.geojson.colorScale('Reds', {
                     min: 0,
                     max: 100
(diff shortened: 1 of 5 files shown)


danlipsa:
@aashish24 @matthewma7 @dorukozturk Please review.

dorukozturk:
@danlipsa The functionality works great. 
There are 2 things:
1) When clicked on the clamp checkbox first time, the input boxes do not appear.
deepinscreenshot20170512081236

2) We should probably limit the clamp functionality to numeric properties only. 

Other than that LGTM.

danlipsa:
@dorukozturk For 1. Do you need to expand the clamping layout? For 2. I agree, even more we need to limit all scaling (linear, log and quantile) to numeric properties only, not only clamping. I plan to address this next time I work on minerva.

dorukozturk:
@danlipsa if we expand the layout by default that should solve the first problem. The second issue is more important I think. Let me know if you need anything from me.

danlipsa:
@dorukozturk As we discussed, the first problem is a feature of how the panels (Clamping, Fill and Stroke) are implemented, not a bug. :smile: If its checkbox is unchecked, a panel is loaded hidden. Maybe a way to 
make it easier for the user would be to expand the panel when the user checks its checkbox.

I filed an issue for the second problem
https://github.com/Kitware/minerva/issues/406

Ready to merge?

PR #408, opened 2017-05-24
Twofishes integration

dorukozturk:
In order to get the twofishes up and running do this:
```sh
$ pip install ansible
$ git clone git@github.com:OpenGeoscience/fsqio.git
$ cd fsqio/ansible
$ ansible-galaxy install -r requirements.txt
$ vagrant up geocoder
```
It takes a while to provision but once it succeeds, it should be up and running on port 8087.
http://localhost:8087 

deepinscreenshot20170525083830


server/rest/twofishes.py
```diff
@@ -0,0 +1,115 @@
...
+from girder.api import access
+from girder.api.describe import Description
+from girder.api.rest import Resource
+from girder.utility.model_importer import ModelImporter
+from girder.plugins.minerva.rest.geojson_dataset import GeojsonDataset
+from girder.plugins.minerva.utility.minerva_utility import findDatasetFolder
+
+
+class TwoFishes(Resource):
+    def __init__(self):
```
aashish24: @dorukozturk doing initial review. Should we add some docs for the class as well?

server/rest/twofishes.py
```diff
@@ -0,0 +1,115 @@
...
+    def postGeojson(self, params):
+        twofishes = params['twofishes']
+        try:
+            locationInfo = json.loads(params['locations'])
+            geojson = TwoFishes.createGeojson(twofishes, locationInfo)
+        except ValueError:
+            locationInfo = params['locations']
+            geojson = TwoFishes.createGeojson(twofishes, locationInfo)
+
+        minervaDataset = self.createMinervaDataset(geojson, params['name'])
```
aashish24: I found it little confusing that post method returns a minerva dataset where as get returns geojson?

server/rest/twofishes.py
dorukozturk: Yes, I will add docstring for this.

server/rest/twofishes.py
dorukozturk: Returning the minerva dataset from the POST request is consistent with Girder. Whenever we create a new item via a POST request we return the item in Girder. I think the idea is if you are creating it, you will use it at some point so you will need it's id. If the relationship between GET and POST is confusing I can change on of the routes to something else.

aashish24: Right, I was thinking if get should also return a dataset as opposed to a geojson? But I do know what is expected from get in Girder so whatever is consistent, I am good with that.

dorukozturk: That endpoint does not have any side effects. It just gets a WKT from twofishes and converts that to geojson. I thought it might be useful since a lot of people work with geojson in BSVE. Also it was good for debugging.

dorukozturk:
Once we merge fsqio PR, I am going to replace the instructions here. The most dramatic difference between master fsqio and our fork is with countries. 

deepinscreenshot20170602100652
deepinscreenshot20170602101059

server/rest/twofishes.py
aashish24: okay sounds good to me. Let me ask @manthey @jbeezley for suggestions here.

aashish24:
> @aashish24 Once we merge fsqio PR, I am going to replace the instructions here. The most dramatic difference between master fsqio and our fork is with countries.

our fork data looks great!

dorukozturk:
@aashish24 This is ready for another review. I added some tests. Coverage for the twofishes.py file is 88%.

server/rest/twofishes.py
```diff
@@ -0,0 +1,142 @@
...
+    def createGeometryFromWkt(wkt):
+        """Creates a shapely geometry from wkt"""
+        return loads(wkt)
+
+    @staticmethod
+    def createGeojson(twofishes, locations):
+        """Create geojson for given locations and twofishes url"""
+
+        try:
+            locations = ast.literal_eval(locations)
```
jbeezley: What is the purpose of the eval here?

dorukozturk: When the user passes a list of inputs through swagger we get the string representation of the list. It is there to construct the list from the string.

PR #411, opened 2017-06-02
Update to the latest version of geojs

jbeezley:
@matthewma7 Could you verify this fixes the issue you reported with styling polygons?


matthewma7:
I just verified it fixed the issue. 👍 
Thank you @jbeezley @manthey for looking into this.

PR #410, opened 2017-06-01
Create dataset from Postgres

matthewma7:
This PR allows users to create datasets from their Postgres tabular data.
Geometry data can be used if geometry columns exist in table.
If not, the tabular data could be linked with existing geojson dataset on the file to get geometry data.

This PR adds https://github.com/OpenGeoscience/girder_db_items as a plugin dependency.

PS: Some predefined GeoJSON dataset is provided for geometry linking. Change extension to .geojson before uploading to Minerva.

chrome_2017-06-01_16-24-41
[world_countries.geojson.txt](https://github.com/Kitware/minerva/files/1045967/world_countries.geojson.txt)
[us_states.geojson.txt](https://github.com/Kitware/minerva/files/1045968/us_states.geojson.txt)
[us_counties.geojson.txt](https://github.com/Kitware/minerva/files/1045969/us_counties.geojson.txt)

PR #416, opened 2017-06-19
Create dataset from Postgres (continue)

matthewma7:
This PR is the continuation of https://github.com/Kitware/minerva/pull/410.
The original PR was accidentally merged and closed. Its changes have been reverted.


server/utility/minerva_utility.py
```diff
@@ -48,6 +48,11 @@ def findMinervaFolder(currentUser, user, create=False):
                            PluginSettings.MINERVA_FOLDER, create)
 
 
+def findPublicFolder(currentUser, user, create=False):
```
dorukozturk: My editor was changing the file modes and it also happened here. Can you change the file mode back to 100644? Or I can do it.

server/rest/postgres_geojson.py
```diff
@@ -0,0 +1,368 @@
...
+                        'table': table,
+                        'schema': schema}],
+            'fields': fields,
+            'group': group,
+            'filters': filters,
+            'limit': -1,
+            'format': output_format
+        }
+
+    # def _getProperties(self, params):
```
dorukozturk: Can we remove this if not needed.

server/utility/minerva_utility.py
matthewma7: Oh, I probably unintentionally changed it. If it's not too much trouble, you could change it back.

dorukozturk: Ok I reverted it back.

server/rest/postgres_geojson.py
```diff
@@ -0,0 +1,368 @@
...
+
+        if geometryField['type'] == 'built-in':
+            buildInGeomField = geometryField['field']
+            properties = [field, {
+                'func': aggregateFunction,
+                'param': {'field': field}
+            }]
+            # add string fields with concat aggregate function and in the format for
+            # json_build_object
+            for property in [(i['name'], {
```
dorukozturk: Can we make this more readable? Maybe extract some of the expressions to separate variables?

matthewma7: Totally, I got confused with the weird indentation myself when I read it just then.

server/rest/postgres_geojson.py
```diff
@@ -0,0 +1,368 @@
...
+            group = [buildInGeomField]
+        elif geometryField['type'] == 'link':
+            fields = [{
+                'func': aggregateFunction,
+                'param': {'field': field},
+                'reference': field
+            }]
+            group = [x['value'] for x in geometryField['links']]
+            # add string fields with concat aggregate function
+            for i in [i for i in self._getColumns(assetstore, {'table': params['table']})
```
dorukozturk: Here you are looping through a list that is created by a list comprehension. I think you don't need the list comprehension here. IMHO you should either do a for loop or list comprehension but not both. I believe this is too complicated to be a list comprehension and will be easier to read if it is just a for loop.

matthewma7: Agreed. Updated.

web_external/js/views/widgets/PostgresWidget.js
```diff
@@ -0,0 +1,473 @@
...
+        'change select#geometry-built-in-field': '_geometryBuiltInFieldChange',
+        'change .geometry-field-type': '_geometryFieldTypeChange',
+        'change select.link-target': '_linkTargetChange',
+        'change select.link-field': '_linkFieldChange',
+        'change select.link-operator': '_linkOperatorChange',
+        'change select.link-value,input.link-value': '_linkValueChange',
+        'click button.add-link': '_addLinkClick',
+        'click button.remove-link': '_removeLinkClick'
+    },
+    formToTable: {
```
manthey: Is this used for anything?

matthewma7: Good catch. No, it's not used. I'll remove it.

PR #418, opened 2017-06-29
Two styling related fixes

matthewma7:
This PR fixes two issues,

1, The previous Postgres feature applies a default style to geometries because we would like to have initial `fillColorKey` value, this also fixes the behavior that the initial styling on the map and the styling on the styling modal doesn't match. However, that change has an issue that the default styling won't be applied at first rendering to geojson that doesn't have style. 

2, The GeojsonStyleWidget had an issue that it won't recognize geojson LineString becuase it is looking for the style `Line` instead of `LineString`.


matthewma7:
This is ready for review.

PR #420, opened 2017-07-03
Dataset download api endpoint

matthewma7:
Minerva core and it's dependent like [gaia minerva](https://github.com/OpenDataAnalytics/gaia_minerva) are using girder file download API endpoint to get the data of Minerva Dataset. 
However, with the introduction of Postgres on-the-fly geometry linking dataset, the existing girder file download API is not supporting downloading of such dataset because the data needs on-the-fly processing.
We should unify the downloading interface of all Minerva dataset. Monkey patching file/download api is cumbersome and error prone. Also, as Minerva dataset are getting more complex, it makes sense to build Minerva Dataset specific API instead of work around girder concepts. 
So, this PR creates minerva_dataset/download API to replace the role of file/download endpoint. This new endpoint output geojson download no matter it is a GeoJSON dataset or Postgres geometry linking dataset.

### DIFF
--- a/server/rest/dataset.py
+++ b/server/rest/dataset.py
@@ -21,12 +21,13 @@ import os
 import shutil
 import pymongo
 import tempfile
+import json
 
 from girder.api import access
-from girder.api.describe import Description
-from girder.api.rest import Resource, loadmodel, RestException
+from girder.api.describe import Description, autoDescribeRoute
+from girder.api.rest import Resource, loadmodel, RestException, GirderException
 from girder.constants import AccessType
-from girder.utility import config
+from girder.utility import config, assetstore_utilities
 
 from girder.plugins.minerva.constants import PluginSettings
 from girder.plugins.minerva.utility.minerva_utility import findDatasetFolder, \
@@ -38,6 +39,7 @@ import girder_client
 
 
 class Dataset(Resource):
+
     def __init__(self):
         self.resourceName = 'minerva_dataset'
         self.route('GET', (), self.listDatasets)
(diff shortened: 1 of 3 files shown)


matthewma7:
This is ready for review.

aashish24:
@matthewma7 is this ready for review?

matthewma7:
@aashish24 Yes, it's ready for review.

server/rest/dataset.py
```diff
@@ -400,3 +403,84 @@ def createGeojson(self, item, params):
...
         .errorResponse('Write permission denied on the Item.', 403))
+
+    @access.public
+    @autoDescribeRoute(
+        Description('Download a dataset.')
+        .modelParam('id', model='item', level=AccessType.READ)
+        .errorResponse('ID was invalid.')
+        .errorResponse('Read access was denied on the parent folder.', 403))
+    def download(self, item, params):
+        minervaMeta = item['meta']['minerva']
```
aashish24: @matthewma7 I guess if in the future we add support for other databased, we will have to add them here?

aashish24: I am wondering if in the future, we can use some sort of registration to register code to download data depending on the data format. The current solution is fine for now.

matthewma7: @aashish24, You are right. It needs some code change, but if additional database support is acquired through https://github.com/OpenGeoscience/database_assetstore, only minor change is needed to generalize different database. I'll keep in mind of this in future development.

matthewma7:
@aashish24 Thanks!

Issue #423, opened 2017-07-06
Reordering layer causes layer to disappear


matthewma7:
Reordering layers on a session unexpectedly cause layers to disappear on the map.

Reproduce steps:
Adding three layers to map, continuously click **move down** on the layer appear first in the list, some layers will become invisible on the map.

PR #425, opened 2017-07-12
Fixes issues with layer reordering

matthewma7:
This addresses the issue https://github.com/Kitware/minerva/issues/423.

This basically implements a new logic for calculation the 'stack' value of datasets when reordering.

Issue reproduce:
Add more than three datasets to session layer, reorder layer freely, the map and panel will start to behave unexpectedly.

Fix result:

Sample datasets:
(Any dataset will do, These three are provided for convinience. Rename .txt to .geojson before uploading)
[buffer.geojson.txt](https://github.com/Kitware/minerva/files/1145108/buffer.geojson.txt)
[intersect.geojson.txt](https://github.com/Kitware/minerva/files/1145109/intersect.geojson.txt)
[within.geojson.txt](https://github.com/Kitware/minerva/files/1145110/within.geojson.txt)

### DIFF
--- a/web_external/views/body/LayersPanel.js
+++ b/web_external/views/body/LayersPanel.js
@@ -70,45 +70,51 @@ const LayersPanel = Panel.extend({
     },
 
     reorderLayer: function (event) {
-        var prevDataset, nextDataset;
-        var datasetId = $(event.currentTarget).attr('m-dataset-id');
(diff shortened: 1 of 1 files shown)


matthewma7:
Screen records added. 🖼

matthewma7:
@dorukozturk Thanks 😃

PR #427, opened 2017-07-19
Makes it ready to enable adding WMS layer

matthewma7:
The adding WMS layer button was disabled, but the functionality is kept in the code.
A code bug issue was introduced during girder 2 migration that preventing enable adding WMS layer.

This PR fixes the issue and make it easy to enable the WMS feature easily.

### DIFF
--- a/web_external/templates/body/dataPanel.pug
+++ b/web_external/templates/body/dataPanel.pug
@@ -2,6 +2,8 @@ include ../layout/panelMixins.pug
 
 +panel-title('Datasets', 'm-data-panel')
   i.m-upload-local.m-icon-enabled.icon-upload(title="Upload local file")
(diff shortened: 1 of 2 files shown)


web_external/templates/body/dataPanel.pug
```diff
@@ -2,6 +2,8 @@ include ../layout/panelMixins.pug
 
 +panel-title('Datasets', 'm-data-panel')
   i.m-upload-local.m-icon-enabled.icon-upload(title="Upload local file")
+  // uncomment this line to enable adding wms layer
```
aashish24: @matthewma7 don't we need this line?

matthewma7:
@aashish24 Thanks. Sorry I missed your comment earlier, I'll merge it now.

Issue #426, opened 2017-07-18
JavaScrit linting (ESLint of tests/externals) are not passing yet


aashish24:

matthewma7:
Related PR has been merged. This can be closed now.

PR #428, opened 2017-07-24
Adds eslint test & fixes eslint errors

matthewma7:
This PR fixes the issue https://github.com/Kitware/minerva/issues/426.
There is some code refactoring applied for fixing the linting errors.


web_external/views/body/DataPanel.js
```diff
@@ -236,7 +238,7 @@ export default Panel.extend({
 
     deleteDatasetEvent: function (event) {
         if ($(event.currentTarget).hasClass('icon-disabled')) {
```
manthey: We should probably invert this conditional so we don't have an empty block and then an else.

web_external/views/widgets/LegendWidget.js
```diff
@@ -1,28 +1,28 @@
-/**
-* This widget displays the list of WMS layers
-*/
-minerva.views.LegendWidget = minerva.View.extend({
+// /**
```
manthey: Why is this file all comments?  Should it be removed or decommented?

matthewma7:
@manthey, Thanks for the review. I confirmed that the LegendWidget is not used and removed it.

PR #431, opened 2017-08-31
Make sure webpack config has expected entries.

manthey:
Also, update setuptools so that pip can install other packages properly.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,7 +28,7 @@ before_install:
     - export CXX=g++-4.8
     - nvm install v6
     - nvm use v6
-    - pip install -U pip virtualenv
+    - pip install -U pip virtualenv setuptools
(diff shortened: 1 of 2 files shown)


matthewma7:
Hi, @manthey, This looks cool. 
I have noticed there is something changed with the girder master but didn't look into it.

This PR fixes the webpack resolve helper Hammer issue. 
However, due to some change of girder master, this command will throw an error

`girder-install web --watch-plugin minerva`

```
Configuring plugin gravatar (plugin.yml, build=OFF)
Loading "plugin.js" tasks...ERROR
>> TypeError: Cannot read property 'module' of undefined
```

Do you have some idea on this error?

jbeezley:
I have PR out in girder to fix that error.

https://github.com/girder/girder/pull/2315

PR #432, opened 2017-09-04
Initial support of ktile raster layer

matthewma7:
This adds initial KTile geotiff layer support to Minerva based on the new [girder_ktile ](https://github.com/OpenGeoscience/girder_ktile) girder plugin.

After adding min and max setting
image


.travis.yml
```diff
@@ -69,6 +69,11 @@ before_install:
     - git clone https://github.com/OpenGeoscience/database_assetstore.git "${DB_ASSETSTORE_DIR}" && git -C "${DB_ASSETSTORE_DIR}" checkout $DATABASE_ASSETSTORE_VERSION
     - pip install -r "${DB_ASSETSTORE_DIR}/requirements.txt" -U
 
+    # install girder_ktile plugin
+    - GIRDER_KTILE_DIR="${GIRDER_SOURCE_DIR}/plugins/girder_ktile"
```
aashish24: I feel that the var name is not representative of the repo. Should we call it GIRDER_DATABASE_ASSETSTORE_DIR

package.json
```diff
@@ -17,7 +17,7 @@
     "jquery-extendext": "^0.1.2",
     "jquery-ui-bundle": "^1.11.4",
     "moment": "^2.17.1",
-    "reconcile.js": "github:matthewma7/reconcile.js"
+    "reconcile.js": "^1.3.2"
```
aashish24: 👍

matthewma7: @aashish24 Yes, I contacted the author and merged in the feature I needed.

dorukozturk:
@matthewma7 You might want to expose minimum and maximum values for the single band styling case through the UI. They default to min:0 and max:256 but that won't apply to every layer.  https://github.com/OpenGeoscience/girder_ktile/blob/master/server/rest.py#L32-L35

matthewma7:
Hi @dorukozturk, is there a way to disable the min and max? or it has to be set? It's kind of hard to give good initial max and min values. Or do you think to set them to 0 and 256 as default is sufficient?

dorukozturk:
@matthewma7 For now we can let the user give it to us. It requires some processing on the back end to get min and max which we have not implemented yet.

server/rest/dataset.py
```diff
@@ -349,6 +349,14 @@ def promoteItemToDataset(self, item, params):
                 minerva_metadata['original_files'] = [{
                     'name': file['name'], '_id': file['_id']}]
                 break
+            elif 'tif' in file['exts'] and file['mimeType'] == 'image/tiff':
```
dorukozturk: @matthewma7 Uploading a regular tiff file as opposed to a geotiff file breaks this. We should be able to check whether the "srs" is empty  or not in the info call (GET /ktile/{id}/info).

.travis.yml
```diff
@@ -65,9 +65,14 @@ before_install:
...
-    - DB_ASSETSTORE_DIR="${GIRDER_SOURCE_DIR}/plugins/database_assetstore"
-    - git clone https://github.com/OpenGeoscience/database_assetstore.git "${DB_ASSETSTORE_DIR}" && git -C "${DB_ASSETSTORE_DIR}" checkout $DATABASE_ASSETSTORE_VERSION
-    - pip install -r "${DB_ASSETSTORE_DIR}/requirements.txt" -U
+    - DATABASE_ASSETSTORE_DIR="${GIRDER_SOURCE_DIR}/plugins/database_assetstore"
+    - git clone https://github.com/OpenGeoscience/database_assetstore.git "${DATABASE_ASSETSTORE_DIR}" && git -C "${DATABASE_ASSETSTORE_DIR}" checkout $DATABASE_ASSETSTORE_VERSION
+    - pip install -r "${DATABASE_ASSETSTORE_DIR}/requirements.txt" -U
+
+    # install girder_ktile plugin
+    - GIRDER_KTILE_DIR="${GIRDER_SOURCE_DIR}/plugins/girder_ktile"
+    - git clone https://github.com/OpenGeoscience/database_assetstore.git "${GIRDER_KTILE_DIR}"
```
manthey: Shouldn't this clone `girder_ktile`, not `database_assetstore`?

matthewma7: @manthey Totally, don't what black magic made the Travis build work..

server/rest/dataset.py
matthewma7: @dorukozturk, Got it, I see what you mean.

PR #433, opened 2017-09-07
Style on layer

matthewma7:
This PR moves the styling option onto layer panel.
Since BSVE doesn't have a good session concept, the style is still saved to the dataset, when needed (with a checkbox).

Also, I talked about a library called reconcile.js, which mimics the way React update the dom. Over the past development with Minerva, I constantly find myself need the DOM merging updating. 
I included it in this PR as a utility function (Probably will always be), it worked great for this PR. I think it will help this and future developments. 
However, it's currently using a forked version of mine, which adds an option to help parent child view relationship.

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,8 @@
     "jade": "~1.11.0",
     "jquery-extendext": "^0.1.2",
     "jquery-ui-bundle": "^1.11.4",
-    "moment": "^2.17.1"
+    "moment": "^2.17.1",
+    "reconcile.js": "github:matthewma7/reconcile.js"
   },
   "devDependencies": {
     "grunt-contrib-watch": "^1.0.0"
--- a/web_external/templates/body/layersPanel.pug
+++ b/web_external/templates/body/layersPanel.pug
@@ -7,16 +7,40 @@ include ../widgets/legendWidget.pug
   .layersList
     ul.datasets
       each dataset, index in datasets
-        li.dataset&attributes({'m-dataset-id': dataset.get('_id')})
+        - var attributes = {'m-dataset-id': dataset.get('_id')}
+        li.dataset&attributes(attributes)
           if dataset.get('geoError')
-            i.icon-attention.m-remove-dataset-from-layer&attributes({'m-dataset-id': dataset.get('_id')})
+            i.icon-attention.m-remove-dataset-from-layer&attributes(attributes)
           else
-            i.icon-eye.m-toggle-dataset&attributes({'m-dataset-id': dataset.get('_id')})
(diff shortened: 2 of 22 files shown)


matthewma7:
@aashish24, could you help me review this PR?

package.json
```diff
@@ -16,7 +16,8 @@
     "jade": "~1.11.0",
     "jquery-extendext": "^0.1.2",
     "jquery-ui-bundle": "^1.11.4",
-    "moment": "^2.17.1"
+    "moment": "^2.17.1",
+    "reconcile.js": "github:matthewma7/reconcile.js"
```
aashish24: @matthewma7 should we think about pushing these changes to upstream?

web_external/templates/body/layersPanel.pug
```diff
@@ -7,16 +7,39 @@ include ../widgets/legendWidget.pug
...
             span= dataset.get('name')
             .m-push-right
+              //- Cog icon for geo rendering configuration
+              if geoRenderType === 'geojson' || geoRenderType === 'geojson-timeseries'
+                - var classes = 'icon-cog icon-button m-configure-geo-render'
+                - classes += (dataset.get('displayed') ? ' icon-disabled dataset-in-session' : ' icon-enabled');
+                - classes += ((!dataset.get('displayed') && dataset.get('geoError')) ? ' m-geo-render-error' : '');
+                i(class=classes)&attributes(attributes)
+
+              //- if geoRenderType === 'wms'
```
aashish24: @matthewma7 perhaps we can add a comment here why this code is commented out?

web_external/views/body/LayersPanel.js
```diff
@@ -56,6 +60,32 @@ const LayersPanel = Panel.extend({
...
+    /**
+     * Display a modal dialog allowing configuration of GeoJs rendering
+     * properties for the selected dataset.
+     */
+    configureGeoRender: function (event) {
+        var datasetId = $(event.currentTarget).attr('m-dataset-id');
+        var dataset = this.collection.get(datasetId);
+        var geoRenderType = dataset.getGeoRenderType();
+
+        console.log(dataset);
```
aashish24: left over?

package.json
matthewma7: Yes. I'll create PR see if the author response at all.
Otherwise, I am thinking we can move this to Kitware organization.

web_external/templates/body/layersPanel.pug
matthewma7: Definitely. I'll add some comment. 
In general, since we are not supporting WMS in Minerva currently, the code for WMS is commented out. So, I moved the code but comment it out.

PR #434, opened 2017-09-11
Dataset Panel ui improvement

matthewma7:
This PR improves the Dataset Panel with indented tree view, better text overflow text ellipsis, and many code logic refactoring.

image

### DIFF
--- a/web_external/views/body/DataPanel.js
+++ b/web_external/views/body/DataPanel.js
@@ -28,41 +28,17 @@ export default Panel.extend({
         'click .source-title': 'toggleSources',
         'click .category-title': 'toggleCategories'
     },
-    _addCategories: function (source, category) {
-        this.visibleMenus[source][category] = true;
-    },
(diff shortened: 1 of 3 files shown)


matthewma7:
This is ready for review

web_external/views/body/DataPanel.js
```diff
@@ -299,47 +275,29 @@ export default Panel.extend({
...
-        });
-    },
-
     getDatasetModel: function () {
         return this.parentView.parentView.model;
     },
 
-    render: function () {
-        var model = this.getDatasetModel();
+    render() {
```
aashish24: is the shorthand cross-browser compatible now?

matthewma7: Babel will handle the transformation to es5 and it will end up like this ```render: function render() { ... }```

aashish24: ah, forgot that we are using babel, thanks.

PR #441, opened 2017-09-22
Color legend for map coloring

matthewma7:
This PR adds the color legend support to Minerva.


matthewma7:
@manthey Thank for reviewing and catching that one bug 👍

PR #443, opened 2017-09-26
Feature info new layout

matthewma7:
This PR changes the feature info popup layout from horizontal table layout to vertical table layout.
I tried to tweak many things and this is what I got.

image


matthewma7:
@manthey, Thanks for the reviewing. 👌

PR #444, opened 2017-09-29
Uses all colors that colorbrewer provides for ktile styling

matthewma7:
This PR switches to use the most color for each series of colorbrewer has to provide finer, but not overwhelming fine detail with customed ktile layer style.

image


matthewma7:
@dorukozturk Thank you 👍

PR #446, opened 2017-10-02
Dataset info new layout

matthewma7:
Uses a table layout to show some meta of dataset instead of showing the raw JSON

chrome_2017-10-30_09-49-55

### DIFF
--- a/web_external/stylesheets/widgets/datasetInfoWidget.styl
+++ b/web_external/stylesheets/widgets/datasetInfoWidget.styl
@@ -1,0 +1,4 @@
+@import './infoTableMixin.styl'
+
+.m-dataset-info-table
+  info-table()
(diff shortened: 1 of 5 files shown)


aashish24:
@matthewma7 this looks good. Few suggestions: 

1. Can we remove all the extra whitespace from the right side.. I do not see that space serving any purpose

2. I would suggest that we make the table 100% width and height at bottom half of the dialog

3. Remove the table borders - all of them. I feel that they are distraction? 

I can be convinced otherwise :) these are merely suggestions.

matthewma7:
@aashish24 I can see your suggestion would make it looks better. I will work on that.

matthewma7:
@aashish24, I pushed a commit base on your suggestion, but the table borders are kept but lightened. I no border make the content less focused.

chrome_2017-10-30_09-49-10

aashish24:
thanks @matthewma7 can we resolve the conflict so that I can test again?

matthewma7:
@aashish24 Thanks your review!

PR #447, opened 2017-10-02
Uses single file upload

matthewma7:
This is reported as an issue. 
The solution right now is disabling multiple file upload.

### DIFF
--- a/web_external/views/body/DataPanel.js
+++ b/web_external/views/body/DataPanel.js
@@ -105,7 +105,8 @@ export default Panel.extend({
             noParent: true,
             title: 'Upload a dataset',
             overrideStart: true,
-            parentView: this.parentView
+            parentView: this.parentView,
+            multiFile: false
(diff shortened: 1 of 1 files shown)


aashish24:
@matthewma7 approved this PR. Are we planning to enable multiple file upload in the future?

matthewma7:
@aashish24 I do see the value of multiple updates. That might need some thinking on how to make it work. For now, this PR would make it works consistently. (before this PR, multiple uploads will result in only first file created as dataset, and other files become orphan files to Minerva).

aashish24:
> @aashish24 I do see the value of multiple updates. That might need some thinking on how to make it work. For now, this PR would make it works consistently. (before this PR, multiple uploads will result in only first file created as dataset, and other files become orphan files to Minerva).

+1. Sounds good. Can we create an issue for the multiple file upload then?

Issue #438, opened 2017-09-22
Detect the most valuable attribute for initial styling


matthewma7:
Current, Minerva uses a constant color for all features when a geojson is upload.
However, the code could detect an attribute that is a good candidate for styling and style with a default color ramp.

PR #448, opened 2017-10-02
Try picking a property for initial styling

matthewma7:
This PR adds a feature to automatically tries to select a property to set default style. https://github.com/Kitware/minerva/issues/438
The priority of the selection is 
wider range sequential > smaller range sequential > more category qualitative > less category quantitative

Before:

Ater:

### DIFF
--- a/web_external/models/DatasetModel.js
+++ b/web_external/models/DatasetModel.js
@@ -48,11 +48,39 @@ const DatasetModel = MinervaModel.extend({
         if (this.getDatasetType().match(/(geo)?json/)) {
             var geoData = geojsonUtil.normalize(this.get('geoData'));
             var visProperties = this.getMinervaMetadata().visProperties;
+            // Apply max and min value as styling color clamp initial value
             if (visProperties && visProperties.polygon && visProperties.polygon.fillColorKey && geoData.summary[visProperties.polygon.fillColorKey]) {
                 visProperties.polygon.maxClamp = geoData.summary[visProperties.polygon.fillColorKey].max;
                 visProperties.polygon.minClamp = geoData.summary[visProperties.polygon.fillColorKey].min;
             }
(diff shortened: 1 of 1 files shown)


web_external/models/DatasetModel.js
```diff
@@ -48,14 +48,42 @@ const DatasetModel = MinervaModel.extend({
...
+                    this.getMinervaMetadata().visProperties.polygon = Object.assign(this.getMinervaMetadata().visProperties.polygon, fillColorProperties);
+                }
+            }
             this.set('geoData', geoData);
         }
     },
 
+    tryGetDefaultFillColor(summary) {
+        var numberFields = _.pairs(summary).filter((pair) => {
+            return !pair[1].values;
```
manthey: Although this may have `values`, that isn't a guarantee that it has `min` and `max`.  I think a better check would be
`return !pair[1].values && isFinite(pair[1].min) && isFinite(pair[1].max) && pair[1].min !== pair[1].max;`

web_external/models/DatasetModel.js
```diff
@@ -48,14 +48,42 @@ const DatasetModel = MinervaModel.extend({
...
+        if (numberFields.length) {
+            return {
+                fillColorKey: _.sortBy(numberFields, (pair) => pair[1].max - pair[1].min).reverse()[0][0],
+                fillRamp: 'Reds'
+            };
+        }
+        var categoricalFields = _.pairs(summary).filter((pair) => {
+            return pair[1].values;
+        });
+        if (categoricalFields.length) {
```
manthey: Should we only do this if there are at least two categories, i.e., `if (categoricalFields.length >= 2)`?

matthewma7:
@manthey These are great suggestions, thank you. I updated the code correspondingly.

PR #449, opened 2017-10-09
Adds ability to remove dataset from session on dataset panel

matthewma7:
This PR also fixes a small issue with layer visibility

### DIFF
--- a/web_external/templates/body/dataPanel.pug
+++ b/web_external/templates/body/dataPanel.pug
@@ -26,13 +26,13 @@ if sourceCategoryDataset
                 .m-control-left
                   //- Globe icon to move dataset into current session
                   if dataset.isGeoRenderable()
-                    - var classes = (dataset.get('displayed') ? 'icon-globe icon-disabled dataset-in-session' : 'icon-globe icon-enabled add-dataset-to-session')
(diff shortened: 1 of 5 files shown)


web_external/templates/body/dataPanel.pug
```diff
@@ -26,13 +26,13 @@ if sourceCategoryDataset
                 .m-control-left
                   //- Globe icon to move dataset into current session
                   if dataset.isGeoRenderable()
-                    - var classes = (dataset.get('displayed') ? 'icon-globe icon-disabled dataset-in-session' : 'icon-globe icon-enabled add-dataset-to-session')
+                    - var classes = (dataset.get('displayed') ? 'icon-cancel-circled2 icon-enabled remove_dataset-from-session' : 'icon-globe icon-enabled add-dataset-to-session')
                     i(title='add to layers', class=classes)&attributes(attributes)
```
manthey: Can you also make it so the tooptip (title) changes from "add to layers" to "Remove from layers".  We should also add the same title to the matching icon in the layer panel.

While you are at it, elsewhere most other titles start with uppercase, so change `add to layers` to `Add to layers`.

matthewma7: Good catch 👌

matthewma7:
@manthey, Thank you for the reviewing.

PR #451, opened 2017-10-17
Add color legend support to ktile raster layer

matthewma7:
And some code refactoring

### DIFF
--- a/web_external/views/adapters/Adapters.js
+++ b/web_external/views/adapters/Adapters.js
@@ -611,9 +611,9 @@ rendering.geo.KtileRepresentation = rendering.geo.defineMapLayer('ktile', functi
         if (_.isEmpty(visProperties)) {
             layer.url((x, y, z) => `${url}/${z}/${x}/${y}`);
         } else {
-            layer.url((x, y, z) => `${url}/${z}/${x}/${y}?
-            palette=${visProperties.palettable}&band=${visProperties.band}&
-            minimum=${visProperties.min}&maximum=${visProperties.max}`);
+            layer.url((x, y, z) => `${url}/${z}/${x}/${y}?` +
+            `palette=${visProperties.palettable}&band=${visProperties.band}&` +
+            `minimum=${visProperties.min}&maximum=${visProperties.max}`);
             var colorLegendCategory = {
                 type: 'discrete',
                 scale: 'linear',
--- a/web_external/views/adapters/Adapters.js
+++ b/web_external/views/adapters/Adapters.js
@@ -611,7 +611,9 @@ rendering.geo.KtileRepresentation = rendering.geo.defineMapLayer('ktile', functi
         if (_.isEmpty(visProperties)) {
             layer.url((x, y, z) => `${url}/${z}/${x}/${y}`);
         } else {
-            layer.url((x, y, z) => `${url}/${z}/${x}/${y}?palette=${visProperties.palettable}&band=${visProperties.band}&minimum=${visProperties.min}&maximum=${visProperties.max}`);
+            layer.url((x, y, z) => `${url}/${z}/${x}/${y}?
+            palette=${visProperties.palettable}&band=${visProperties.band}&
+            minimum=${visProperties.min}&maximum=${visProperties.max}`);
             var colorLegendCategory = {
                 type: 'discrete',
                 scale: 'linear',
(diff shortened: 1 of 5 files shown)


manthey:
Should there be a legend even when the colors aren't custom?  Also, it seems like there should be a way to modify the default -- for instance, if I want a black-to-white continuous scale with different min and max values.

aashish24:
@matthewma7 do you have an answer for @manthey's question above?

matthewma7:
@aashish24 I meant to ask @dorukozturk when I got a chance.
I guess I can ask him now.
Hi @dorukozturk, I don't you may not know the answer fully, but since you have more experience with Mapnik than I do, I am wondering that is the default Mapnik render behavior without style predictable? I mean, when there are three bands, it will be rendered to color. When there is, in this case, one band, does it always use the same greyscale? Is the greyscale ordinal or continuous?

dorukozturk:
@matthewma7 [Here](https://github.com/mapnik/mapnik/wiki/RasterColorizer) they list bunch of options. I think your options are 
- mapnik2.COLORIZER_LINEAR 
- mapnik2.COLORIZER_DISCRETE
- mapnik2.COLORIZER_EXACT
- mapnik2.COLORIZER_INHERIT

I think greyscale is the default like you said.

aashish24:
> When there is, in this case, one band, does it always use the same greyscale? Is the greyscale ordinal or continuous

greyscale is assumed when there is only one band. You can look into QGIS which is the industry standard tool. 

greyscale would be continuous (think of height map)

matthewma7:
After a discussion with @aashish24, we decided to implement legend for default raster styling in the separate PR, because it might involve something with some complexity outside the current changes.
An issue has been created for that request, https://github.com/Kitware/minerva/issues/459

web_external/views/adapters/Adapters.js
```diff
@@ -600,21 +601,34 @@ rendering.geo.WmsRepresentation = rendering.geo.defineMapLayer('wms', function (
...
         var fileId = dataset.get('meta').minerva.original_files[0]._id;
-        var visProperties = dataset.getMinervaMetadata().visProperties;
         var url = getApiRoot() + '/ktile/' + fileId;
-        if (visProperties) {
-            layer.url((x, y, z) => `${url}/${z}/${x}/${y}?palette=${visProperties.colorbrewer}&band=${visProperties.band}&minimum=${visProperties.min}&maximum=${visProperties.max}`);
-        } else {
+        if (_.isEmpty(visProperties)) {
             layer.url((x, y, z) => `${url}/${z}/${x}/${y}`);
+        } else {
+            layer.url((x, y, z) => `${url}/${z}/${x}/${y}?palette=${visProperties.palettable}&band=${visProperties.band}&minimum=${visProperties.min}&maximum=${visProperties.max}`);
```
aashish24: possible to wrap this long line?

aashish24:
LGTM if @manthey also approves it.

web_external/views/adapters/Adapters.js
```diff
@@ -600,21 +601,36 @@ rendering.geo.WmsRepresentation = rendering.geo.defineMapLayer('wms', function (
...
         var url = getApiRoot() + '/ktile/' + fileId;
-        if (visProperties) {
-            layer.url((x, y, z) => `${url}/${z}/${x}/${y}?palette=${visProperties.colorbrewer}&band=${visProperties.band}&minimum=${visProperties.min}&maximum=${visProperties.max}`);
-        } else {
+        if (_.isEmpty(visProperties)) {
             layer.url((x, y, z) => `${url}/${z}/${x}/${y}`);
+        } else {
+            layer.url((x, y, z) => `${url}/${z}/${x}/${y}?
+            palette=${visProperties.palettable}&band=${visProperties.band}&
+            minimum=${visProperties.min}&maximum=${visProperties.max}`);
```
manthey: Unbreak this line or change how it is generated.  The extra spaces and newlines mess up the query parameters.  This should work instead:
```
            layer.url((x, y, z) => `${url}/${z}/${x}/${y}?` +
            `palette=${visProperties.palettable}&band=${visProperties.band}&` +
            `minimum=${visProperties.min}&maximum=${visProperties.max}`);
```

matthewma7: Oh, yes. I was carried away with the multiline support of template literals.

PR #452, opened 2017-10-24
Fix the alert pop up style

matthewma7:
The background of the alert popup was transparent, it is because the wrong type is used.
This will make it light red translucent, same as it is in girder.


matthewma7:
Thanks

PR #453, opened 2017-10-24
Dataset sharing

matthewma7:
This PR adds dataset sharing based on Girder public folder structure.


matthewma7:
This is ready for some review.

aashish24:
@matthewma7 how do I test it? Create another user? Also, it would be great if we can add some tests as well (I am okay with if that goes in another PR)

web_external/views/body/DataPanel.js
```diff
@@ -281,24 +372,64 @@ export default Panel.extend({
...
+        if (e.target.checked) {
+            this.selectedDatasetId.add(datasetId);
+        } else {
+            this.selectedDatasetId.delete(datasetId);
+        }
+        this.render();
+    },
+
+    selectCategory(e) {
+        var datasetIds = $(e.target).closest('.m-sub-category').find('.dataset').map((i, el) => $(el).attr('m-dataset-id')).toArray();
```
aashish24: can we wrap this long line? I do not have any strong preference but something < 120 works most of the time. Does Girder has any guideline?

manthey: Girder has requirements for Python line lengths (100 characters), but not for Javascript.  There are endless debates about line lengths.  I personally wish we *wouldn't* break conceptual lines -- let your editor format it visually to whatever you find handy, rather than forcing this into the code; keeping it together makes it more easily found using a tool like grep.  However, others feel strongly otherwise.  There isn't any correct answer to this.

matthewma7: I think in this particular case we could break the line at map() since sometimes the map could have a full function as parameter which would normally on a different line.
PS: I personally always use automatic line wrap in my editor, and the editor always starts the wrapped lines at the proper position (with the same amount of intention as the original line). So I never felt the need to explicitly make a line shorter.

matthewma7:
@aashish24 Yes, I don't see roadblock to for testing this feature. It is the time to familiarize myself with girder testing infrastructure. I think we could create a subsequent PR for it.
To test this PR, you could create a different user B, once the user A shares a dataset, it should be available to user B.

aashish24:
> @aashish24 Yes, I don't see roadblock to for testing this feature. It is the time to familiarize myself with girder testing infrastructure. I think we could create a subsequent PR for it.

+1 lets do that in a separate PR

aashish24:
> To test this PR, you could create a different user B, once the user A shares a dataset, it should be available to user B.

okay will do that

matthewma7:
@aashish24 Thank you for reviewing this PR. 👌

PR #454, opened 2017-10-26
Update the sizing of feature info window

matthewma7:
Use auto size option to have the popup window wrap the content, but has a minimum width

Before:
chrome_2017-10-26_08-28-19

After:
chrome_2017-10-26_08-27-26


matthewma7:
@aashish24 Thanks the reviewing.

PR #457, opened 2017-11-07
Added conda instructions and environment file

aashish24:


aashish24:
@matthewma7 @manthey please review.

jbeezley:
The encryption key is used to store user credentials for external WMS servers.  It avoids having to store those credentials in plain text in the database.  Generating that config option should probably be added somehow to the production deployment.

It is a little tricky because you don't want to change an existing encryption key if you are reusing an existing mongo database that already contains credentials.  I think it should come from some private external setting or file that is used by a specific deployment rather than just generated randomly.

aashish24:
thanks @matthewma7 so I guess that was the missing piece. Can I remove instructions for running pip install -r requirements.txt then?

matthewma7:
@jbeezley Thank you for explaining the encryption.

matthewma7:
@aashish24 Yes, the instructions **pip install -r requirements.txt** can be removed.

aashish24:
@matthewma7 this is ready for your review now.

docs/installation.rst
```diff
@@ -45,54 +45,106 @@ Fedora 22
 .. _installing node.js on Red Hat: https://nodejs.org/en/download/package-manager/#enterprise-linux-and-fedora
 
 
+Conda
+^^^^^^^^^
```
matthewma7: It's working fine but probably should remove the extra characters.

aashish24:
@matthewma7 I forced-push in this case so rst for conda is good now.

docs/installation.rst
```diff
@@ -45,54 +45,106 @@ Fedora 22
...
-- Install girder_db_items plugin into the Girder plugins directory.
+- Install girder_ktile plugin.
 
 ::
 
-   cd GIRDER_DIR/plugins
-   git clone https://github.com/OpenGeoscience/girder_db_items database_assetstore
+    git clone https://github.com/OpenGeoscience/girder_ktile
+    girder-install plugin -s girder_ktile
+    girder-install web --dev --plugins girder_ktile
```
matthewma7: Technically, when we build Minerva, it's dependencies (girder_ktile and database_assetstore) will be built too. So, do you think we need this step?

docs/installation.rst
```diff
@@ -45,54 +45,106 @@ Fedora 22
...
+    git clone https://github.com/Kitware/minerva.git
+    cd minerva
+    conda env create -f conda_env.yml python=2.7
+    source activate minerva-dev
+
+- Run mongo daemon
+
+::
+
+    mongod &
```
matthewma7: With a clean Ubuntu, I need to create a directory for mongo frist then run
**mkdir mongo_data
mongod --dbpath mongo_data &**

docs/installation.rst
```diff
@@ -136,6 +191,12 @@ Notes:
     cd GIRDER_DIR
```
matthewma7: Do we want to remove this alternative method?

docs/installation.rst
```diff
@@ -45,54 +45,106 @@ Fedora 22
...
+Conda installation and conda environment setup should be the first step in
+creating a development environment for Minerva.
+
+- Install Conda (https://conda.io/miniconda.html)
+
+- Create minerva development environment
+
+::
+
+    mkdir geoviz
```
matthewma7: Is this geoviz directory necessary?

docs/installation.rst
```diff
@@ -45,54 +45,106 @@ Fedora 22
...
+::
+
+    mongod &
+
+- Install Girder
+
+::
+
+    cd ..
+    git clone https://github.com/girder/girder.git
```
matthewma7: Should we checkout to girder 2.x-maintenance branch?

docs/installation.rst
```diff
@@ -112,6 +164,9 @@ Notes:
     cd GIRDER_DIR
```
matthewma7: Should we remove this method?

matthewma7:
Hi @aashish24 I added more comment, some are not preventing the installing. 
It's not related to this PR, but some wording and structure can be improved, I'll create a PR for further improvement. 
Overall, this document worked for me to create the environment for Minerva.

docs/installation.rst
aashish24: that is a good point does it work all the time? I always had this issue where in some cases it did not. Did you experience the same?

docs/installation.rst
aashish24: Fixed

docs/installation.rst
matthewma7: Yes, I think it always works for me. In fact, when I am testing this, I only run the Minerva one, and it worked good.

aashish24:
thanks @matthewma7

Issue #463, opened 2017-11-16
Remove unused source models


aashish24:
I believe after our refactoring, we are not longer using the source models (https://github.com/Kitware/minerva/blob/master/web_external/models/SourceModel.js). If that is the case can we remove it @matthewma7 ?

Issue #461, opened 2017-11-16
Remove romanesco code 


aashish24:
@matthewma7 are we still using this code? https://github.com/Kitware/minerva/tree/master/server/jobs 

As we are preparing to use the latest girder worker, if we are not using this code, can we remove it?

PR #464, opened 2017-11-16
Remove unused and abandoned code

aashish24:


aashish24:
Fixes #463 #461

aashish24:
@matthewma7 the WMS portion of code is not clear to me. I think WMS code needs some decent fixing. I am wondering if we should remove the code now and add it back later if we need it. I do not think currently we need it for any active project. thoughts? I or you may have to push more code to remove traces of WMS

matthewma7:
@aashish24 I am not really certain about the WMS code. I would like to ask, @jbeezley, does the WMS imported dataset in BSVE use any of these codes?

jbeezley:
BSVE reference layers derive from WMS layers, but I don't think it relies on this specific code, which is just the "Add a WMS layer" button.

aashish24:
Thanks @jbeezley @matthewma7

PR #467, opened 2017-11-16
Removed some more unused code

aashish24:
@matthewma7 I found some more code which we do not need I think. Can you confirm?


matthewma7:
Yes, these are not used.

aashish24:
thanks if you can approve it, then I can merge it (I took out WIP).

aashish24:
thanks @matthewma7

PR #468, opened 2017-11-21
Show datasets bounding box

matthewma7:
This PR implements a new feature that shows boundaries of selected datasets. 
It uses colors from colorbrewer.pastel1 series for box filling and geojs AnnotationLayer for labeling.

I feel this feature along with previous features regarding datapanel change is asking for more expressive icon usage, specifically [stacked icon](http://fontawesome.io/examples/#stacked). 
The base girderfontello doesn't support stacked icon, but I think I'd introduce stacked icon to minerva in later PR.

This PR also moves the two buttons on Dataset panel title onto action panel.


plugin.cmake
```diff
@@ -1,3 +1,5 @@
+set( ENV{PYTHONPATH} ${CMAKE_CURRENT_LIST_DIR} )
```
manthey: Adding this makes other cmake tests for girder plugin fail, since they prefer the minerva path to their own.

matthewma7: Oh, I see. You probably see the reason I need this is for the https://github.com/Kitware/minerva/pull/468/files/1ce8f9f9d83968119215d0933908d4b104708081#diff-c6aae2354f158fa61778727be393104dR25
which will prepare my data for testing. 
So, at this moment, what I can think of is, 
1, omit this line in this file and ask the developer to set it before running cmake.
2, make a change to girder to make it actually try pickup the TEST_MODULE at root girder and all plugins directories?
3, Will set it first and then unset it after work?

manthey: I've made a PR on Girder (https://github.com/girder/girder/pull/2535) to resolve this.  Assuming that gets merged, we can remove the `set` line and add `TEST_PYTHONPATH "${CMAKE_CURRENT_LIST_DIR}"` to the `minervaSpec` test.

matthewma7: Oh, I see the PR. That's great. Thanks!

requirements.txt
```diff
@@ -9,4 +9,4 @@ matplotlib==1.5.3
 numpy>=1.10.1
 owslib==0.9.1
 python-dateutil==2.4.2
-Shapely==1.6.0
+Shapely==1.6.2
```
manthey: Unless there is a reason to pin to a very specific version, I think all of the entries in requirements.txt should be rewritten as `>=` rather than `==`.

server/rest/dataset.py
```diff
@@ -21,7 +21,10 @@
 import shutil
 import pymongo
 import tempfile
-import json
+import json as Json
```
manthey: This is a little confusing in the actual code, since `json` is the standard module name.  It would be better to rename variables rather than change the module (or import it as `json_module` to be more explicit about why it was done).

server/rest/dataset.py
```diff
@@ -393,7 +397,8 @@ def promoteItemToDataset(self, item, params):
                 minerva_metadata['original_files'] = [{
                     'name': file['name'], '_id': file['_id']}]
                 break
-            elif 'tif' in file['exts'] and file['mimeType'] == 'image/tiff':
+            elif (('tif' in file['exts'] or 'tiff' in file['exts'])
```
manthey: Another way to write this which would extend to more extensions would be `{'tif', 'tiff'}.intersection(file['exts']) and file['mimeType'] == 'image/tiff')`.

server/rest/dataset.py
```diff
@@ -526,16 +547,62 @@ def _getPostgresGeojsonData(self, item):
...
 
+            return Geojson.FeatureCollection(assembled)
+
+    @access.public
+    @autoDescribeRoute(
+        Description('Calculate bounding box of a dataset.')
+        .modelParam('id', model='item', level=AccessType.READ)
+        .errorResponse('ID was invalid.')
+        .errorResponse('Read access was denied on the parent folder.', 403))
+    def getBound(self, item, params):
```
manthey: You may want to add server tests for `getBound` and `download`.  They get tested via the client tests, but aren't reported as covered.

web_external/views/map/MapPanel.js
```diff
@@ -283,9 +286,75 @@ const MapPanel = Panel.extend({
...
+                        [bounds.ulx, bounds.uly],
+                        [bounds.lrx, bounds.uly],
+                        [bounds.lrx, bounds.lry]
+                    ]]
+                },
+                properties: {
+                    annotationType: 'rectangle',
+                    name: dataset.get('name'),
+                    labelFontWeight: 'normal',
+                    labelColor: 'grey',
```
manthey: I find the `grey` has low contrast on many of the background colors.  Perhaps the label should be darker with a light shadow to help highlight it?

matthewma7:
Hi @manthey, thanks for your review suggestions. I have made the changes accordingly.

PR #472, opened 2017-12-04
Update geojson-timeseries data summary calculation

matthewma7:
Previously the dataset is using the summary of only the first frames in the series

### DIFF
--- a/web_external/geojsonUtil.js
+++ b/web_external/geojsonUtil.js
@@ -122,9 +122,12 @@ geojsonUtil.normalize = function normalize(geojson) {  // eslint-disable-line co
                 var label = '' + (entry.label || (entry.time ? moment(entry.time).format('L LTS') : null) || ('Frame ' + (normalized.series.length + 1)));
                 var time = moment.utc(entry.time);
                 normalized.series.push({ time: time, geojson: norm, label: label });
-                $.extend(normalized.summary, norm.summary);
             }
         });
+        normalized.summary = geojsonUtil.accumulate(
+            _.flatten(normalized.series.map((series) => series.geojson.features))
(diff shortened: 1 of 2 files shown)


matthewma7:
@manthey I think the current logic won't get the **summary** all value range of all frames. With this PR it's now getting the summary of all frames, but could you review the logic and see if it make sense?

manthey:
I'm not sure what the right logic is, but this is probably a strict improvement.

Imagine that you have two frames, each with three elements with a scalar and a categorical value.  Frame 0 has `[{s: 2, c: 'a'}, {s: 3, c: 'b'}, {s: 4, c: 'c'}]` and frame 1 has `[{s: 2, c: 'a'}, {s: 3, c: 'b'}, {s: 5, c: 'd'}]`  (only the last element has changed).  Before, I think we ended up with a summary from frame 1: `{s: {count: 3, min: 2, max: 5, nFinite: 3, sum: 9, sumsq: 38}, c: {count: 3, values: {a: 1, b: 1, d: 1}}}`.  With this, we get `{s: {count: 6, min: 2, max: 5, nFinite: 6, sum: 19, sumsq: 67}, c: {count: 6, values: {a: 2, b: 2, c: 1, d: 1}}}`

For categorical values, combining them makes sense.  For scalar values, we want the minimum and maximum values, but I don't know what we want as far as sums (I'm not sure where those are used).

matthewma7:
I see. 
Based on my memory and a quick search in the source code, it seems we are not using the sums at all at this moment.

manthey:
In that case, this is a good change.  If we ever use them, we may need to revisit what we do.

matthewma7:
@manthey Thank you for the reviewing. 👍

PR #473, opened 2017-12-06
Fix coverage reporting

manthey:
This properly tests coverage in travis.

It reports coverage of both python and javascript to Codecov.

It reports javascript coverage to cdash.  If we are happy with Codecov's reporting, then it may not be worth the effort to get python coverage in cdash.


manthey:
Once this gets merged to master, we can add a codecov badge, too.

matthewma7:
@manthey This works great for my [PR](https://github.com/Kitware/minerva/pull/468).

PR #474, opened 2017-12-08
Add tutorial to docs

dorukozturk:
Adds a getting started section to the documentation.


dorukozturk:
@aashish24 PTAL.

matthewma7:
Had a discussion with @dorukozturk. Would apply some changes based on my suggestions.
1, Move GeoJSON section to the front of GeoTiff
2, Change some usage of the word Should to Would
3, Simplify the paragraph about Geotiff default styling and technical detail.

PR #479, opened 2017-12-14
Draw boundary dataset and boundary filtering

matthewma7:
Utilize GeoJS annotation layer for boundary dataset creating. 
Uses bounding box support based on the previous PRs for intersecting filtering.

### DIFF
--- a/web_external/views/map/MapPanel.js
+++ b/web_external/views/map/MapPanel.js
@@ -327,7 +327,7 @@ const MapPanel = Panel.extend({
                 });
                 layer.geoOn(geo.event.annotation.state, (e) => {
                     bootbox.prompt({
-                        title: 'Give dataset a name?',
+                        title: 'Boundary dataset name',
                         value: 'Boundary',
                         callback: (name) => {
                             if (name !== null) {
--- a/web_external/views/map/MapPanel.js
+++ b/web_external/views/map/MapPanel.js
@@ -1,6 +1,7 @@
 import _ from 'underscore';
 import geo from 'geojs';
 import colorbrewer from 'colorbrewer';
+import bootbox from 'bootbox';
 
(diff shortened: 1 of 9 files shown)


manthey:
When you delete a boundary that was used to filter, should the filter be turned off?

manthey:
When you are in draw-boundary mode, can you change the draw icon so that it shows it is selected?

web_external/views/map/MapPanel.js
```diff
@@ -318,11 +319,34 @@ const MapPanel = Panel.extend({
...
 
+        this.listenTo(events, 'm:draw-boundary-dataset', () => {
+            if (!this.drawDatasetLayer) {
+                var layer = this.map.createLayer('annotation', {
+                    annotations: ['rectangle'],
+                    showLabels: false
+                });
+                layer.geoOn(geo.event.annotation.state, (e) => {
+                    bootbox.prompt({
+                        title: 'Give dataset a name?',
```
manthey: I would change this title to `Boundary dataset name` (or at least remove the question mark).

matthewma7: Make sense.

matthewma7:
@manthey Thank you for your review. I think it's ok not to remove the boundary when the dataset is removed? 
I think the icon could have changed to indicate it's drawing mode, but I did some new feature related to icons. So, I think I will add it in a following PR. 

When testing, I found out the boundary calculation is not supported on WMS layers, so I added a series of check and message to handle the situation when boundary can't be acquired.

manthey:
If it wouldn't be too much work, can you add a test with a dataset where you can't get bounds?  Otherwise, this looks good.

matthewma7:
Thank you @manthey for reviewing 🍺

PR #480, opened 2017-12-22
Gui improvement

matthewma7:
See commits message for detail.


web_external/stylesheets/body/mapPanel.styl
```diff
@@ -5,6 +5,13 @@
 .m-map-panel-map
   width 100%
   height 100%
+  
+  &:focus
+    outline none
+
+// Remove geojs map focus highlight
+.m-map-panel-map.geojs-map.highlight-focus:after
```
manthey: Another way to accomplish this is to ask geojs to not use the focus highlight.  This could be done immediately after creating the map in MapPanel.js:
```
this.map = geo.map(...);
var interactorOpts = this.map.interactor().options();
interactorOpts.keyboard.focusHighlight = false;
this.map.interactor().options(interactorOpts);
```
If the intent is to disable keyboard interaction with the map, that can be done, too.  The focus highlight is intended to let keyboard users know when they are operating on the map.

PR #481, opened 2018-01-09
Gaia and girder-worker integration

matthewma7:
Gaia was a third party dependency for Minerva. 
Before this PR, Gaia can be used in Minerva with the plugin gaia-minerva and based on an analysis registering mechanism. 
This PR integrates Gaia-minerva into Minerva, making gaia a dependancy of Minerva. 
This PR also introduce girder-worker to this project. Girder-worker runs Gaia analysis process remotely. 

This PR utilize selection feature of datapanel. This PR applies default values as gaia parameter based on user selection. 

Note: Some of the gaia process is resulting in some errors but it's consistent with the previous state, so I would like to ask to look into those in a separate PR. 
The Minerva server has an endpoint (originally from gaia-minerva) based on some introspection of gaia, so the server became depending on Gaia as well.


aashish24:
@matthewma7 is this ready for review?

matthewma7:
@aashish24 Yes, I think @jbeezley may have looked at it already. 
Since the corresponding change in minerva-docker depends on this. I think we could possibly merge this one first.

jbeezley:
I'm looking at it now.

gaia_tasks/__init__.py
```diff
@@ -0,0 +1,20 @@
...
+__email__ = 'kitware@kitware.com'
+__version__ = '0.0.0'
+
+
+from girder_worker import GirderWorkerPluginABC
+
+
+class GaiaTasks(GirderWorkerPluginABC):
+    def __init__(self, app, *args, **kwargs):
+        self.app = app
```
jbeezley: This function isn't needed.

gaia_tasks/inputs.py
```diff
@@ -0,0 +1,187 @@
...
+
+import json
+import os
+import tempfile
+from base64 import b64encode
+
+import fiona
+import geopandas
+import requests
+# from girder.plugins.minerva.utility.minerva_utility import decryptCredentials
```
jbeezley: remove

gaia_tasks/inputs.py
```diff
@@ -0,0 +1,187 @@
...
+        if uri:
+            self.uri = uri
+        else:
+            tmpdir = tempfile.mkdtemp()
+            self.uri = tempfile.mkstemp(suffix='.json', dir=tmpdir)[1]
+        self.filename = name
+        girderHost = config.getConfig().get('minerva', {}).get('worker_girder_host_name', 'localhost')
+        girderScheme = config.getConfig().get('minerva', {}).get('worker_girder_host_scheme', 'http')
+
+        girderPort = config.getConfig()['server.socket_port']
```
jbeezley: If I understand the code correctly, this will run in worker, which doesn't necessarily run on the same machine as girder.  You have to pass this information from the producer side (girder).

girder_worker will bind a correct girder client instance to your task for you to use.  See for example [this integration test](https://github.com/girder/girder_worker/blob/3c6b4774f8477d4107e3667885f64b901f90eb02/tests/integration/common_tasks/common_tasks/test_tasks/girder_client.py).

gaia_tasks/inputs.py
```diff
@@ -0,0 +1,187 @@
...
+        self.client.addMetadataToItem(self.id, item_meta)
+        return os.path.join(
+            self.client.urlBase, 'file', upload['_id'], 'download')
+
+    def filter_data(self):
+        """
+        Apply filters to the dataset
+        :return:
+        """
+        self.data = filter_pandas(self.data, self.filters)
```
jbeezley: In the spirit of new style girder_worker tasks, this class should really be implemented as a transform like those defined in [girder_io.py](https://github.com/girder/girder_worker_utils/blob/master/girder_worker_utils/transforms/girder_io.py).  Much of what you are doing here is already implemented in the existing transforms.  I don't have a clear sense at the moment whether or not this is an easy thing to do (or even makes any sense) given the current API in Gaia.

server/rest/gaia/analysis.py
```diff
@@ -0,0 +1,84 @@
...
+
+    @access.user
+    @autoDescribeRoute(
+        Description('Run a Gaia analysis.')
+        .jsonParam('analysis',
+                   'JSON describing the output dataset name and Gaia process',
+                   paramType='body')
+    )
+    def gaiaAnalysisTask(self, analysis, params):
+        print analysis
```
jbeezley: remove

server/rest/gaia/analysis.py
```diff
@@ -0,0 +1,84 @@
...
+            'dataset': dataset,
+            'analysis': gaia_json,
+            'token': token
+        }
+        result = gaia_task.delay(kwargs, girder_job_title=datasetName)
+        job = result.job
+
+        addJobOutput(job, dataset)
+
+        # r = result.get()
```
jbeezley: remove

gaia_tasks/tasks.py
```diff
@@ -0,0 +1,35 @@
+import json
+import re
+from girder_worker.app import app
+from girder_worker.utils import girder_job
+from gaia.parser import deserialize
+
+
+@girder_job()
+@app.task(bind=True)
+def gaia_task(self, kwargs):
```
kotfic: Choosing to implement a single task in this way will seriously degrade the capabilities of Gaia with regard to Girder Worker.  As written the current task implements input and output concepts that are more appropriately handled as apart of celery's underlying canvas architecture (e.g. chaining,  groups,  map-reduce style chords).  The purpose of a girder worker task is the encapsulate a function that does some unit of actual work.  As written this encapsulates a function that encapsulates functions confusing the spheres of responsibility. 

It is difficult to foresee how this design choice will limit your ability to use upstream features of Girder Worker.

gaia_tasks/inputs.py
matthewma7: I felt the way I am providing the config is unorthodox, but the current implementation of girder.utility.config is able to get the config no matter when it's running as girder or girder_worker.
But I do see the Bind injected grider_client is better. I'll change the code to use that. Do you know if there is a sample code that I can refer to on "passing this information from the producer side"

kotfic: I think @jbeezley  point was,  the config may not be available because girder worker may (and in fact will likely) be running on a totally different computer.  

In general when developing with girder worker it is a good idea to at least be running girder worker in a separate virtualenv from girder to help surface some (although not all) of these issues

matthewma7: @kotfic, Thank you for your comments, totally get what you and @jbeezley are saying. At this moment I copied the config to the worker instance/container, so it would work. However, I am aware this is not how we should use the new girder-worker. 
As people discussed in Gumbo meeting, I'll learn the proper way to use girder worker (with another project) and come back to this.

jbeezley: After our conversation, it occurred to me that it would be better and easier to understand if you deployed this information into the [worker configuration file](https://github.com/girder/girder_worker/blob/master/girder_worker/worker.dist.cfg) instead.  You should be able to get the config object at `girder_worker.config`.

gaia_tasks/__init__.py
matthewma7: It seems without the __init__, girder-worker won't be able to pick up the task.

gaia_tasks/inputs.py
matthewma7: @jbeezley I see. I pushed a commit to use girder worker config instead. I also pushed a commit to the related PR in minerva-docker.

matthewma7:
Hi @jbeezley, I updated code to use girder-worker config, could you take another look?

matthewma7:
Thanks for the reviewing and feedbacks 👍 @jbeezley

PR #483, opened 2018-01-10
Change the Dataset sharing to based on User Group instead of Public folder

matthewma7:
This PR adds two dependencies to the autojoin and user group.
Based on current observation, the undergoing auto provisioning could include these new dependencies.

### DIFF
--- a/server/utility/minerva_utility.py
+++ b/server/utility/minerva_utility.py
@@ -19,16 +19,17 @@
 from cryptography.fernet import Fernet
 from girder.utility import config
 from girder.utility.model_importer import ModelImporter
+from girder.exceptions import AccessException
 
 from girder.plugins.minerva.constants import PluginSettings
 
 
 def findNamedFolder(currentUser, user, parent, parentType, name, create=False,
-                    public=False):
+                    joinShareGroup=None, public=False):
     folders = \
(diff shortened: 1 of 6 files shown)


server/utility/minerva_utility.py
```diff
@@ -75,11 +81,23 @@ def findDatasetFolder(currentUser, user, create=False):
...
+    if not datasetSharingGroup:
+        raise AccessException('user group "{0}" doesn\'t exist'.format(
+            PluginSettings.DATASET_SHARING_GROUP_NAME))
+    if datasetSharingGroup['_id'] not in currentUser['groups']:
+        raise AccessException('user doesn\'t belong to user group "{0}"'.format(
+            PluginSettings.DATASET_SHARING_GROUP_NAME))
+
     folders = folderModel.find({
-        'public': True,
+        'public': False,
```
manthey: Do we need to insist that this *isn't* a public folder.  It seems like we should no longer care if the folder is public or not in this method.

matthewma7: You are right. No need to limit to private folders when looking up shared datasets. Code updated.

matthewma7:
@manthey Thanks for your reviewing 👌

PR #485, opened 2018-01-30
Dataset name text filtering

matthewma7:

### DIFF
--- a/web_external/views/body/DataPanel.js
+++ b/web_external/views/body/DataPanel.js
@@ -41,7 +41,8 @@ export default Panel.extend({
         'click .action-bar button.remove-bounds': 'removeBounds',
         'click .action-bar button.toggle-bounds-label': 'toggleBoundsLabel',
         'click .action-bar button.intersect-filter': 'intersectFilter',
-        'click .action-bar button.remove-filter': 'removeFilter'
+        'click .action-bar button.clear-filters': 'clearFilters',
+        'keyup .search-bar input': 'applyNameFilter'
     },
 
     toggleCategories: function (event) {
@@ -299,6 +300,8 @@ export default Panel.extend({
         this.showSharedDatasets = !!this.sessionModel.getValue('showSharedDatasets');
         this.selectedDatasetsId = new Set();
         this.filters = {};
+        this.nameFilterKeyword = '';
+        this.applyNameFilter = _.debounce(this.applyNameFilter, 300);
         this.listenTo(this.collection, 'g:changed', function () {
             this.render();
         }, this).listenTo(this.collection, 'change', function () {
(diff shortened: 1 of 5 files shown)


web_external/views/body/DataPanel.js
```diff
@@ -507,6 +514,33 @@ export default Panel.extend({
...
+        this.nameFilterKeyword = keyword;
+        if (keyword) {
+            this.filters.name = this.collection.models
+                .reduce((ids, dataset) => {
+                    var match = false;
+                    try {
+                        var regex = new RegExp(keyword, 'i');
+                        match = !!regex.exec(dataset.get('name'));
+                    } catch (ex) {
+                        match = dataset.get('name').toLocaleLowerCase().indexOf(keyword) !== -1;
```
manthey: I recommend moving this below of the catch as `match |= dataset.get('name').toLocaleLowerCase().indexOf(keyword) !== -1;`.  Otherwise, if I have a dataset called "foo.bar" and search for "foo.b", it won't be found (because as a regex that doesn't match), which would be surprising to a user.

manthey: To be clearer:
```
                    var match = false;
                    try {
                        var regex = new RegExp(keyword, 'i');
                        match = !!regex.exec(dataset.get('name'));
                    } catch (ex) {}
                    match |= dataset.get('name').toLocaleLowerCase().indexOf(keyword) !== -1;
                    if (match) {
                        ids.push(dataset.get('_id'));
                    }
                    return ids;
```

aashish24:
@matthewma7 @manthey as discussed in the meeting, we would like to extend it to search through the metadata as well. Are we planning on doing that in this branch or another one?

matthewma7:
@aashish24 I personally is thinking to do it in another PR because we kind want this to release soon, and I also have a concern that searching meta may end up with too many records, so some mechanism to indicate might be needed (a checkbox, a sorted result, etc)

aashish24:
> @aashish24 I personally is thinking to do it in another PR because we kind want this to release soon, and I also have a concern that searching meta may end up with too many records, so some mechanism to indicate might be needed (a checkbox, a sorted result, etc)

Sounds good. We should start tagging releases (thoughts?)

aashish24:
@matthewma7 @manthey how do we enable coverage reporting on PR's?

manthey:
Girder changed where coverage files are stored.  I just made PR #487 to fix the reporting.

web_external/views/map/MapPanel.js
```diff
@@ -336,6 +336,7 @@ const MapPanel = Panel.extend({
                             layer.removeAllAnnotations();
                         }
                     });
+                    prompt.find('.bootbox-body input').attr('maxlength', '30');
```
manthey: Out of curiosity, why 30?  I know wiredTiger supports up to 1023 bytes "including structural overhead"; probably half that would be pretty safe.

matthewma7: I might be wrong. I was thinking other characters from other language takes more bytes than English letters, so I choose it to make it safe. But if 30 is too limiting, I think we change it to a bigger number. (A failed creating boundary is not that devastating)

manthey: Sure -- I think wiredTiger uses UTF-8, which would mean that *in theory* a codepoint could take 6 bytes, but in practice 4 is sufficient.  How about we make it 250?  It is large enough most people shouldn't have an issue and small enough Mongo shouldn't have an issue.  I only discovered this because I intentionally paste absurd data into forms.

matthewma7: Sounds good. New commit pushed.

matthewma7:
@manthey Thank you for the review all the edge case you found.

manthey:
:+1:

PR #493, opened 2018-03-06
Map screenshot and map scale widget

matthewma7:
Implement taking map screenshot feature with a map ui layer widget

### DIFF
--- a/plugin_tests/client/minervaSpec.js
+++ b/plugin_tests/client/minervaSpec.js
@@ -42,7 +42,7 @@ describe('Main view', function () {
                     var existingCreateWidget = layer.createWidget;
                     layer.createWidget = function (widgetName) {
                         if (widgetName === 'slider') {
-                            return {};
+                            return { canvas: function () { } };
                         } else {
                             return existingCreateWidget.apply(this, arguments);
                         }
@@ -350,5 +350,21 @@ describe('Layerpanel', function () {
         var zoom = mapPanel.map.zoom();
         $('.layersList .dataset').first().find('.m-zoom-to').trigger('click');
         expect(zoom).not.toEqual(mapPanel.map.zoom());
+    });
+});
+
+describe('Mappanel', function () {
+    it('Take screenshot', function () {
+        $('button.m-screenshot').click();
+        waitsFor(function () {
+            return $('.m-screenshot-result').length;
+        }, 'screenshot be taken and shown');
+        runs(function () {
+            var screenshotResult = $('.m-screenshot-result');
+            var image = screenshotResult.find('img').attr('src');
+            var image2 = screenshotResult.find('a.save').attr('href');
+            expect(image).toEqual(image2);
+            expect(image.startsWith('data:image/png;base64,')).toBe(true);
+        });
     });
 });
--- a/plugin_tests/client/minervaSpec.js
+++ b/plugin_tests/client/minervaSpec.js
@@ -29,38 +29,6 @@ $(function () {
 });
 
 describe('Main view', function () {
-    var _map;
-    beforeEach(function () {
-        // disable the ui slider because it doesn't work on phantomjs
-        _map = geo.map;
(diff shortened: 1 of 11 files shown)


web_external/views/map/MapPanel.js
```diff
@@ -418,6 +430,20 @@ const MapPanel = Panel.extend({
             trigger: 'hover'
         };
         this.$('.m-save-current-baselayer').tooltip(tooltipProperties);
+
+        var screenshotButtonContainer = $(this.screenshotWidget.canvas());
+        screenshotButtonContainer.empty().append(screenshotButtonTemplate());
+        screenshotButtonContainer.find('button').on('click', () => {
+            var layers = this.map.layers().filter((layer) => layer.layerName !== 'ui');
+            this.map.screenshot(layers, { wait: 'idle' }).then((image) => {
```
manthey: The `map.screenshot` function should either be `map.screenshot(layers, undefined, undefined, {wait: 'idle'})` or `map.screenshot({layers: layers, wait: 'idle'})`.

aashish24:
@matthewma7 is this ready for review? Looks like travis is failing.

matthewma7:
@aashish24 David and I looked into why it's failing. David is fixing the issue in Girder and GeoJS. I'll change this to [WIP]

manthey:
Fundamentally, there is a [webkit bug](https://bugs.webkit.org/show_bug.cgi?id=17352) (also see [here](https://bugs.webkit.org/show_bug.cgi?id=29305) and [here](https://bugs.webkit.org/show_bug.cgi?id=129172))  that prevents webkit-based browsers from using the canvas to take a screenshot of HTML components.  This applies to PhantomJS (which we use for testing), iOS, and probably Safari.

There is a minor issue in the GeoJS screenshot code in that if a css file is missing, the screenshot fails.  I'll have a PR for that soon.  But this doesn't fix the underlying webit issue.  For webkit browsers, the best we can do with a screenshot is get the canvas elements (webGL and 2d canvas layers), and skip the html and svg layers (though some svg would be possible with a bunch of work).

I think the best thing to do would be the first time a screenshot is attempted that wants an HTML layer, we test if such a thing is supported.  If not, we emit a console warning and then skip rendering such layers.  This has the virtue that webkit browsers would still get screenshots of base maps and many features (and, thus, not be completely broken).  The Minerva screenshot test would succeed (though would not contain the scale in the PhantomJS test).  Unless there is some reason not to, I'll make another GeoJS PR to this effect.

matthewma7:
Updated to latest GeoJS and fixed the build error. Thanks @manthey 

Also, added scale widget to the map.
image

plugin_tests/client/minervaSpec.js
```diff
@@ -42,7 +42,7 @@ describe('Main view', function () {
                     var existingCreateWidget = layer.createWidget;
                     layer.createWidget = function (widgetName) {
                         if (widgetName === 'slider') {
-                            return {};
+                            return { canvas: function () { } };
```
manthey: At least on the latest GeoJS, I don't think we need this `beforeEach` and `afterEach`.  At least, locally, the tests all pass without these.

web_external/views/map/MapPanel.js
```diff
@@ -130,13 +138,25 @@ const MapPanel = Panel.extend({
                     ? mapSettings.basemap_args
                     : {});
             this.uiLayer = this.map.createLayer('ui');
-            this.uiLayer.createWidget('slider', { position: { right: 20, bottom: 30 } });
+            this.uiLayer2 = this.map.createLayer('ui');
```
manthey: This would be clearer with a more explicit name (`controlLayer` perhaps?) and a comment saying that widgets on this layer won't be included in the screenshot.

manthey:
@matthewma7 This is failing due to pip 10.0's release.  I'll make a PR to address that soon.

PR #494, opened 2018-03-12
Some improvements

matthewma7:
Please find commit message for detail


web_external/models/DatasetModel.js
```diff
@@ -327,6 +327,15 @@ const DatasetModel = MinervaModel.extend({
         this.set('visible', true);
         this.set('opacity', 1);
         this.set('displayed', false);
+    },
+
+    getBoundSupported() {
+        if (this.get('_id').toString().length < 7) {
```
manthey: Where is an example of one of these short ids?

matthewma7: Oh, thanks for your suggestion, I made a change with the following commit.

PR #495, opened 2018-03-16
Rename Dataset feature

matthewma7:


aashish24:
@matthewma7 I see lot of white-space in the popup in my testing. Is that something we can fix. I think anytime we have have a very large white space, I prefer we fit it nicely so that we do not hide the background.

matthewma7:
@aashish24 I see. I'll update this one's style.

matthewma7:
@aashish24 I updated the style to avoid whitespace. Please take another look.

matthewma7:
@aashish24 Thanks for the reviewing. 👍

PR #498, opened 2018-03-27
Persist in memory dataset

matthewma7:
Minerva is able to receive in-memory vector dataset, however, since it's in memory, server-side functionality can't be used. This PR provide a feature with a button to persist an in-memory dataset. Once an in-memory dataset is preserved, it behaves the same as file upload dataset.


manthey:
This seems to work for me:
```
        this._createDatasetfromGeometry(dataset.get('name'), geometry)
              .then((newDataset) => {
                  if (dataset.get('displayed')) {
                      dataset.removeFromSession();
                      this.collection.once('add', (addDataset) => {
                          addDataset.set('displayed', true);
                      });
                  }
                  this.collection.remove(dataset);
                  this.collection.add(newDataset);
              });
```
where I handle showing the new dataset on the backbone add callback.

This isn't quite right -- It moves the dataset to the top of the layers when it is readded.

Also, I've been injecting a memory dataset via the console using a command
```
minerva.events.trigger('m:addExternalGeoJSON', {name: 'Test 1', data: {type: 'FeatureCollection', features: [{type: 'Feature', properties: {}, geometry: {type: 'Point', coordinates: [-91.75, 41.25]}}]}});
```
and I note that it has different stroke color than its setting show.  I think this is an orthogonal issue.

PR #499, opened 2018-03-30
Specify Gaia explicitly from github.

manthey:
Update how setup.py parses requirements.txt.


manthey:
It would be better not to use dependency links since pip says it is deprecated (see https://github.com/pypa/pip/issues/4187), but I'm not sure if there is an alternative (direct urls might work, but when I try adding `gaia@<giit hub url>`, I get a `Direct url requirement are not allowed for dependencies` error from pip 10).

PR #502, opened 2018-04-12
Infoviz widget

matthewma7:
This PR implements an InfovizWidget to visualize a dataset. The requires a vector dataset with a visualizing numeric property, otherwise, a soft error message will be thrown.


jeffbaumes:
Cool. The transparency makes things look cluttered to me. Not sure if having it in a transparent state when not focused is a great feature. Otherwise this looks nice.

matthewma7:
@jeffbaumes I see. And transparency is kind of out of fashion nowadays. 
I think I'll remove the transparency for now. @Everyone
image

manthey:
If I pick two datasets, I find it confusing that the two panels appear directly atop one another.  Perhaps we should stagger them down by roughly the height of the title bar?

Also, should picking a dataset that is already be shown do anything?  Like bring that window to the top?

web_external/views/body/LayersPanel.js
```diff
@@ -32,16 +32,18 @@ const LayersPanel = Panel.extend({
...
-            a.click();
-            document.body.removeChild(a);
-        }).loadGeoData();
+        dataset
+            .loadGeoData()
+            .then(() => {
+                var data = dataset.get('geoData');
+                var a = window.document.createElement('a');
+                a.href = window.URL.createObjectURL(new Blob([JSON.stringify(data)], { type: 'application/json' }));
+                var filename = dataset.get('name') + '.geojson';
```
manthey: A minor improvement (which could be in a different PR): I have some datasets I uploaded called `<name>.geojson`, when I download the, I end up with a file called `<name>.geojson.geojson`.  If the dataset name ends with `.geojson`, we shouldn't append a second `.geojson` to its name.

matthewma7:
@manthey Those are great suggestions. I implemented them, please take a look at the latest commit. 

Preview:
image

manthey:
The staggered position is a lot better.  What determines the z-order of new panels?  If I open multiple panels, then close some of them, then reopen the ones I closed, sometimes the reopened panels appear behind the existing panels.

When opening multiple panels, should their initial position be related to their position in the datasets list?  I think currently they are positioned in the order that the data is finished fetching.

matthewma7:
The z-order is controlled by jquery Dialog, but I think there are ways to tell it to do what you described, I'll take a look later.

matthewma7:
Hi @manthey, I implemented the way you suggested, please take another look.

matthewma7:
@manthey Thank you for your review of all recent PRs. 👍

PR #506, opened 2018-04-24
Fix double .geojson extension when downloading dataset

matthewma7:
This PR fix the issue when downloaded dataset could have double .geojson extension.


manthey:
`_getPostgresGeojsonData()` does return a function where the geometry is in the same databasebase as the other information.

